PR #15, opened 2016-07-27
Upgragde to Checkstyle 7.0

marschall:
Upgrades the checkstyle version from 6.12.1 to 7.0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,9 @@ Official announcement of project transfer - https://groups.google.com/d/topic/so
 
 This plugin provides coding rules from [Checkstyle](http://checkstyle.sourceforge.net/).
 
-Checkstyle Plugin|2.0|2.1.1|2.2|2.3|2.4
------------------|---|---|---|---|---
-Checkstyle|5.6|5.6|6.1|6.4.1|6.12.1
+Checkstyle Plugin|2.0|2.1.1|2.2|2.3|2.4|2.5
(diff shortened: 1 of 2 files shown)


tsjensen:
Don't you need to update the [rules.xml](https://github.com/checkstyle/sonar-checkstyle/blob/master/sonar-checkstyle-plugin/src/main/resources/org/sonar/plugins/checkstyle/rules.xml) also?

marschall:
You're correct. I will do this.

romani:
It would be better to make few middle versions for version between 6.12 - 6.19, to let use that versions for clients.
it is very beneficial to use the same version of CS in IDE  and build process and in Sonar.

tsjensen:
Well we should be grateful for any version! It is a lot of work to update everything to a new version. If he does it for 7.0, that's fantastic.
In the long run, we should think of ways to avoid this chore.

romani:
I am suggesting gradual update(for each release) of metadata that will be beneficial for all.
Gradual to make it simple to update, final update for metadata will be the same. The simpler update the faster it could be done.

romani:
>In the long run, we should think of ways to avoid this chore.

Yes, such problem have all plugins, we started discussion about this with author of eclipse-cs in our maillist, but we do not have that much members in team to make it happen. Everyone is welcome to contribute.

marschall:
I'll do a new PR for 7.1

PR #17, opened 2016-08-07
Add missing 7.1 translations

marschall:
Add missing translations for 7.1.
Also update the parent version of the submodule POMs.


marschall:
It is important that the build still fails because of a conflicting Guava version of Checkstyle 7 (19) and Sonar (10).

The only way to fix this issue would be to bump the sonar API to 6.0. But then the DefaultFileSystem related tests would have to be changed so they compile again.

romani:
we need to exclude guava from dependencies and include version that sonar need.
update for api 6.0 might be a big deal.

PR #18, opened 2016-08-10
Fix master branch build

marschall:
This commit contains a couple of changes that fix the master branch
build:

 - Add Sqale configurations for new rules where it makes sense. I'm not
   familiar with Sqale and had to guess.
 - Set a classifier for the shaded artifact and use that. This makes
   sure the relocated Guava version is used for Checkstyle and the
   not-relocated one for Sonar.
 - Incremented the expected rule count.
 - Add more mission translations.

I did not yet test a SNAPSHOT build against Sonar.

### DIFF
--- a/checkstyle-all/pom.xml
+++ b/checkstyle-all/pom.xml
@@ -46,6 +46,7 @@
               <goal>shade</goal>
             </goals>
             <configuration>
+              <shadedArtifactAttached>true</shadedArtifactAttached>
               <dependencyReducedPomLocation>${project.build.directory}/dependency-reduced-pom.xml</dependencyReducedPomLocation>
               <artifactSet>
                 <includes>
--- a/sonar-checkstyle-plugin/pom.xml
+++ b/sonar-checkstyle-plugin/pom.xml
@@ -83,6 +83,7 @@
       <groupId>${project.groupId}</groupId>
       <artifactId>checkstyle-all</artifactId>
       <version>${project.version}</version>
+      <classifier>shaded</classifier>
     </dependency>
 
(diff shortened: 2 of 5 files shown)


romani:
Please share a command that you use that is failed , we need to use it in travis

marschall:
I used `mvn clean test` the default of TravisCI. An argument could be made for `mvn clean package` because then additionally sonar-packaging-maven-plugin does its thing.

romani:
thanks a lot for a fix, sorry for delay with merge.

romani:
@marschall , I setup travis and now it has problems in "verify" stage , https://travis-ci.org/checkstyle/sonar-checkstyle/jobs/156307077

can you help us with this  ?

marschall:
I'll have a look at it tomorrow.

marschall:
It's a case of too beaucoup
```
[WARNING] Rule 0: org.apache.maven.plugins.enforcer.RequireFilesSize failed with message:
/home/travis/build/checkstyle/sonar-checkstyle/sonar-checkstyle-plugin/target/sonar-checkstyle-plugin-3.1-SNAPSHOT.jar size (7155145) too large. Max. is 5500000/home/travis/build/checkstyle/sonar-checkstyle/sonar-checkstyle-plugin/target/sonar-checkstyle-plugin-3.1-SNAPSHOT.jar
```

Likely caused by the fact that Checkstyle 7.1 plus dependencies is simply bigger now.

From the commit log it looks like this has already been an issue several times in the past, for example when upgrading to 6.9 (SONARCHKST-36).

I'll open a PR that sets more generous bounds.

Issue #29, opened 2016-12-13
rules.xml have 150 items but checks in checkstyle 153


romani:
http://checkstyle.sourceforge.net/checks.html - 153 items
rules.xml - 150 items

3 items are missed.
What about filters/suppressions ?

romani:
now it is 154  in rules.xml vs 153 of checks.html
due to FileContentsHolder - https://github.com/checkstyle/sonar-checkstyle/blob/master/sonar-checkstyle-plugin/src/main/resources/org/sonar/plugins/checkstyle/rules.xml#L1975

it was added recently: https://github.com/checkstyle/sonar-checkstyle/commit/9d217e3372821b96c6b077b96c9568ded2a9f034
It is visible in Sonar Rules: http://localhost:9000/coding_rules#q=File%20Contents%20Holder

romani:
> What about filters/suppressions ?

suppression are configurable in separate place.
![sonar-checkstyle-suppressions](https://cloud.githubusercontent.com/assets/812984/21277491/74474090-c38b-11e6-8541-35edbee0a73f.png)

romani:
all is done, issue is closed

Issue #28, opened 2016-12-13
Update for a lot of Check to make their Sonar config up to implementation, update for rules.xml file (UT validation)


romani:
https://github.com/checkstyle/sonar-checkstyle/blob/master/sonar-checkstyle-plugin/src/main/resources/org/sonar/plugins/checkstyle/rules.xml

There is also https://github.com/checkstyle/sonar-checkstyle/blob/master/sonar-checkstyle-plugin/src/main/resources/org/sonar/l10n/checkstyle.properties with a lot of propetries info

one more registry of Checks:
https://github.com/checkstyle/sonar-checkstyle/blob/master/sonar-checkstyle-plugin/src/main/resources/org/sonar/l10n/checkstyle.propertiestchecks base on groups of Sonar.

registry of documentations for all Checks:
https://github.com/checkstyle/sonar-checkstyle/tree/master/sonar-checkstyle-plugin/src/main/resources/org/sonar/l10n/checkstyle/rules/checkstyle

in runtime we could load all checks and by reflection match all items in this xml

Attention:
validation is missed that problem - https://github.com/checkstyle/sonar-checkstyle/commit/b5b3d1f0c10401ac25780ad8653c9cc02712aa53

romani:
list of available tokens is still not validated, not violation after update to checkstyle 7.4 for NoWhitespaceAfter

romani:
Interesting project https://github.com/startupheroes/startupheroes-checkstyle/tree/master/startupheroes-sonar-file-generator-maven-plugin , unfortunately we can not use/copy their code but we can try to use them as binary

ozlmulg:
@romani We updated our license file [(startupheroes-checkstyle)](https://github.com/startupheroes/startupheroes-checkstyle) You can use our codes if you want. Good works.

rnveach:
`checkstyle.propertiestchecks` needs to be validated for tokens, as it contains a duplicate of acceptable and default tokens. It should also be validated for default value for other properties as they contain that too.

rnveach:
from my previous comment. Sonar displays default values in a gray font below the property description if the `rules.xml` specifies something other than an empty string.
We probably shouldn't display the defaults in `checkstyle.propertiestchecks` if sonar will display the gray text for us.
You can see this for `tokens` in `SuppressWarningsCheck` and `IllegalTokenTextCheck`.

rnveach:
[...]

PR #31, opened 2016-12-13
Issue #28: added UT for validating rules and fixed any errors

rnveach:
Issue #28

This now validates rules.
Guava is needed for `com.google.common.reflect.ClassPath`.
Added 4 missing checks which should fix part of #29 .


romani:
commit is merged to resolve problems in configs.
but tests are marked as Ignore and need to be re-implemented.

PR #50, opened 2016-12-23
Issue #28: fix broken rules test

rnveach:
Issue #28

Ant dependency is because classpath search was throwing class not found exception without it. I assume it is because we have ant classes in checkstyle main code.

Test is fixed, updated with new code from checkstyle.
Test was tweaked to skip filters. I am leaving fileset checks in until we can ascertain if they should be in or not.

### DIFF
--- a/checkstyle-sonar-plugin/pom.xml
+++ b/checkstyle-sonar-plugin/pom.xml
@@ -94,6 +94,13 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>org.apache.ant</groupId>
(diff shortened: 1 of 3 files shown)


rnveach:
Internal test ran fine.

https://travis-ci.org/checkstyle/sonar-checkstyle/jobs/186458317#L297-L298
> Running org.sonar.plugins.checkstyle.internal.ChecksTest
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.112 sec - in org.sonar.plugins.checkstyle.internal.ChecksTest

PR #66, opened 2017-02-08
Lessen dependency on Guava

marschall:
The plugin currently depends on Guava. For checkstyle itself this is
not an issue but for the plugin it has negative consequences. Before
SonarQube 5.2 this meant that the version of Guava the plugin uses and
the one that SonarQube uses had to match. This can be a challenge
because the Google versioning policy is rather opionated. In SonarQube
5.2 and later sonar-packaging-maven-plugin stubbornly refuses to
package Guava because it is not aware of the class loading changes in
5.2 and assumes Guava is still provided by SonarQube when in fact it's
not.

With these changes the only Guava dependencies left in the source of
the plugin are VisibleForTesting. As this is only an annotation for
documentation purposes it is inconsequential if it can't be loaded at
runtime.

The tests still depend on Guava but this is fine.

 [1] https://docs.sonarqube.org/display/DEV/Build

### DIFF
--- a/checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleConfiguration.java
+++ b/checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleConfiguration.java
@@ -26,6 +26,7 @@ import java.io.OutputStreamWriter;
 import java.io.Writer;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Properties;
 
@@ -39,7 +40,6 @@ import org.sonar.api.config.Settings;
 import org.sonar.api.profiles.RulesProfile;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.collect.ImmutableList;
 import com.puppycrawl.tools.checkstyle.ConfigurationLoader;
 import com.puppycrawl.tools.checkstyle.DefaultConfiguration;
 import com.puppycrawl.tools.checkstyle.PropertiesExpander;
@@ -86,7 +86,11 @@ public class CheckstyleConfiguration implements BatchExtension {
     Iterable<File> files = fileSystem.files(predicates.and(
       predicates.hasLanguage(CheckstyleConstants.JAVA_KEY),
       predicates.hasType(InputFile.Type.MAIN)));
-    return ImmutableList.<File>builder().addAll(files).build();
+    List<File> fileList = new ArrayList<>();
+    for (File file : files) {
+      fileList.add(file);
+    }
+    return fileList;
   }
 
   public File getTargetXmlReport() {
--- a/checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleExecutor.java
+++ b/checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleExecutor.java
@@ -19,12 +19,15 @@
  */
 package org.sonar.plugins.checkstyle;
 
+import java.io.Closeable;
 import java.io.File;
+import java.io.IOException;
 import java.io.OutputStream;
 import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URL;
 import java.net.URLClassLoader;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Locale;
@@ -37,8 +40,6 @@ import org.sonar.api.utils.TimeProfiler;
 import org.sonar.plugins.java.api.JavaResourceLocator;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.collect.Lists;
-import com.google.common.io.Closeables;
 import com.puppycrawl.tools.checkstyle.Checker;
 import com.puppycrawl.tools.checkstyle.PackageNamesLoader;
 import com.puppycrawl.tools.checkstyle.XMLLogger;
@@ -61,14 +62,24 @@ public class CheckstyleExecutor implements BatchExtension {
    * Execute Checkstyle and return the generated XML report.
    */
   public void execute() {
-    TimeProfiler profiler =
-            new TimeProfiler().start("Execute Checkstyle " + CheckstyleVersion.getVersion());
-    ClassLoader initialClassLoader = Thread.currentThread().getContextClassLoader();
-    Thread.currentThread().setContextClassLoader(PackageNamesLoader.class.getClassLoader());
-    URLClassLoader projectClassloader = createClassloader();
 
     Locale initialLocale = Locale.getDefault();
     Locale.setDefault(Locale.ENGLISH);
+    ClassLoader initialClassLoader = Thread.currentThread().getContextClassLoader();
+    Thread.currentThread().setContextClassLoader(PackageNamesLoader.class.getClassLoader());
+    URLClassLoader projectClassloader = createClassloader();
+    try {
+      executeWithClassLoader(projectClassloader);
+    } finally {
+      Thread.currentThread().setContextClassLoader(initialClassLoader);
+      Locale.setDefault(initialLocale);
+      close(projectClassloader);
+    }
+  }
+
+  private void executeWithClassLoader(URLClassLoader projectClassloader) {
+    TimeProfiler profiler =
+            new TimeProfiler().start("Execute Checkstyle " + CheckstyleVersion.getVersion());
     Checker checker = new Checker();
     OutputStream xmlOutput = null;
     try {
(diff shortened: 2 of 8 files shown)


rnveach:
CI is failing because of checkstyle violations.
https://travis-ci.org/checkstyle/sonar-checkstyle/jobs/199547421#L367
> [INFO] There are 2 errors reported by Checkstyle 7.3 with https://raw.githubusercontent.com/checkstyle/checkstyle/checkstyle-7.3/config/checkstyle_checks.xml ruleset.
[ERROR] src/main/java/org/sonar/plugins/checkstyle/CheckstyleExecutor.java:[97,11] (coding) NestedTryDepth: Nested try depth is 2 (max allowed is 1).
[ERROR] src/main/java/org/sonar/plugins/checkstyle/CheckstyleSensorException.java:[17] (regexp) RegexpMultiline: Unnecessary consecutive lines

IntelliJ has 2 new violations:
> CheckstyleProfileExporter.java (1)
96: appendCheckerModules() Iteration over activeRulesByConfigKey.keySet() may be replaced with 'entrySet()' iteration
CheckstyleSensorException.java (1)
6: CheckstyleSensorException CheckstyleSensorException does not define a 'serialVersionUID' field

marschall:
Checks are passing now.

romani:
@marschall , please describe a problem that you see and why you want this update.

I like the idea in general but I need more details on problems that users has.

Currently sonar plugin supports Sonar 5.6.4 as minimal version - https://github.com/checkstyle/sonar-checkstyle/blob/master/README.md , why you mention 5.2 ?

> With these changes the only Guava dependencies left in the source of
the plugin are VisibleForTesting

I am fine to remove this annotation completely from code, it is useless. Method should be hidden from outside, if UT need it, it should use reflection to get access to it.

=====================

Please update commit message to reference this PR ("Pull #66: ...........") we need to keep reference to whole discussion.

checkstyle-sonar-plugin/pom.xml
```diff
@@ -330,6 +330,11 @@
                 <branchRate>100</branchRate>
                 <lineRate>78</lineRate>
               </regex>
+              <regex>
+                <pattern>org.sonar.plugins.checkstyle.CheckstyleExecutor</pattern>
+                <branchRate>90</branchRate>
+                <lineRate>94</lineRate>
+              </regex>
```
romani: Sorry, this is not accepted.
All new code should have 100% coverage, to prove that it does not have any dead code. 
Yes, we are not ideal now, as we got huge technical dept from previous owner.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleConfiguration.java
```diff
@@ -86,7 +86,11 @@ public File getXmlDefinitionFile() {
     Iterable<File> files = fileSystem.files(predicates.and(
       predicates.hasLanguage(CheckstyleConstants.JAVA_KEY),
       predicates.hasType(InputFile.Type.MAIN)));
-    return ImmutableList.<File>builder().addAll(files).build();
+    List<File> fileList = new ArrayList<>();
+    for (File file : files) {
+      fileList.add(file);
+    }
+    return fileList;
```
romani: why not to keep original intend as use unmodifiable collection ?

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstylePlugin.java
```diff
@@ -27,15 +28,13 @@
...
-import com.google.common.collect.ImmutableList;
-
 public final class CheckstylePlugin extends SonarPlugin {
 
   private static final String CHECKSTYLE_SUB_CATEGORY_NAME = "Checkstyle";
 
   @Override
   public List getExtensions() {
-    return ImmutableList.of(
+    return Arrays.asList(
```
romani: unmodifiable collection

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleSensor.java
```diff
@@ -49,8 +47,10 @@ public boolean shouldExecuteOnProject(Project project) {
     Iterable<File> mainFiles = fs.files(predicates.and(
       predicates.hasLanguage(CheckstyleConstants.JAVA_KEY),
       predicates.hasType(Type.MAIN)));
-    return !Iterables.isEmpty(mainFiles) &&
-        !profile.getActiveRulesByRepository(CheckstyleConstants.REPOSITORY_KEY).isEmpty();
+    if (!mainFiles.iterator().hasNext()) {
+      return false;
+    }
+    return !profile.getActiveRulesByRepository(CheckstyleConstants.REPOSITORY_KEY).isEmpty();
```
romani: please do single return point from method.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleExecutor.java
```diff
@@ -61,14 +61,24 @@ public CheckstyleExecutor(CheckstyleConfiguration configuration, CheckstyleAudit
...
-    URLClassLoader projectClassloader = createClassloader();
 
     Locale initialLocale = Locale.getDefault();
     Locale.setDefault(Locale.ENGLISH);
+    ClassLoader initialClassLoader = Thread.currentThread().getContextClassLoader();
+    Thread.currentThread().setContextClassLoader(PackageNamesLoader.class.getClassLoader());
+    try (URLClassLoader projectClassloader = createClassloader()) {
+      executeWithClassLoader(projectClassloader);
+    } catch (IOException e) {
+      throw new CheckstyleSensorException("could not close classloader", e);
```
romani: we can not be sure that here will be only problem with closing of classloader. you run some "executeWithClassLoader" that by name do a lot of stuff.
Why you need special exception type for this ?

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleConfiguration.java
marschall: The ownership of the list transfers from the callee to the caller. Modifying the list does not modify or even corrupt the state of the callee. The same is true for the list below.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleExecutor.java
marschall: `executeWithClassLoader` does not throw a checked exception and neither does `createClassloader`. So the `IOException` can only come from the `close` method of `URLClassLoader`. The method is ultimately called by `org.sonar.api.batch.Sensor.analyse(Project, SensorContext)` which does not declare a checked exception so we have to do something here. Other options include:

1. One option would be to simply log a warning. Since the analysis was successful and only the closing of the class loader failed it can be argued that the analysis should complete successfully.
1. An other option is to use an existing unchecked exception or even `RuntimeException` directly, `CheckstyleException` is a checked exception.

In have no problem with either of these.

checkstyle-sonar-plugin/pom.xml
marschall: That would mean removing Java 7 try-with-resources. try-with-resources generates a lot of hidden branches (null checks, suppressed exceptions, …) that make it almost impossible to reach 100% coverage.

I have no problem making this change, I just wanted to point this out.

romani: Yes, I know this problem, I did not recheck your coverage problem.
We experience this problem a lot even in main project - checkstyle. We keep code in old style while this is a problem in cobertura. Benefit that we receive by 100% code coverage is enormous so it do make sense.

Please downgrade your code to try-catch format. You can find examples on how to cover code of try-catch in main repo (powermock, ..... ) .

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleConfiguration.java
romani: agree

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstylePlugin.java
romani: explained.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleExecutor.java
romani: lets go with option 2 and please use IllegalStateException .

romani:
@marschall , CI need to be green.

romani:
@marschall  , please rebase to let me merge.

Issue #45, opened 2016-12-18
activate Checkstyle and sevntu checks over sonar-checkstyle repository


romani:
set of rules could be taken from checkstyle repo.

rnveach:
@romani A lot of the violations have been fixed, but I can't do more without diving into the code for Javadocs, illegal catches, and import control.

muhlba91:
@romani what's the status of this? i somehow see #217 as a follow up?

romani:
Checkstyle is activated to keep code on leash.
Sevntu Checks can be applied in separate issue

Issue #60, opened 2017-01-07
use the same license header but for("LGPLv3 or later") and header as checkstyle project


romani:
request to allow was sent to original authors - Sonar.
https://groups.google.com/forum/#!topic/sonarqube/rQxboJy0nY0

expected header of all files should be: https://github.com/checkstyle/checkstyle/blob/master/config/java.header
Should be enforced for all files by checkstyle validation.

rnveach:
Header was applied in https://github.com/checkstyle/sonar-checkstyle/pull/73

romani:
Sonar team want to keep LGPLv3

>
```
On 9 January 2017 at 12:57, Olivier Gaudin <olivier.gaudin@sonarsource.com> wrote:
Hi Roman,

The LGPL v3 was chosen on purpose for the SonarQube project to offer maximum protection to the community while being business friendly. If there is a specific reason to change I am happy to consider it but otherwise I would tend to push back on this. 

Thanks

Olivier


On Sat, Jan 7, 2017 at 7:04 PM Roman Ivanov <romanivanovjr@gmail.com> wrote:
Hi Sonar team,

During checkstyle plugin transition from Sonarqube to Checkstyle organization
https://groups.google.com/forum/#!topic/sonarqube/HXXxOWS_sOs/discussion

we did not noticed that whole code is under "LGPLv3 or (at your option) any later version"
Example: https://github.com/checkstyle/sonar-checkstyle/blob/master/checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleAuditListener.java

Please confirm that it is ok to convert this code to "LGPLv2.1 or (at your option) any later version"
Header will be: https://github.com/checkstyle/checkstyle/blob/master/config/java.header

https://www.gnu.org/licenses/gpl-faq.html#AllCompatibility
https://www.quora.com/What-are-the-major-differences-between-GNU-LGPL-v3-and-v2-1

So change is very cosmetic, we just allow to allow to use it from earlier version (v2.1).
I am not a lawyer, so I just want to be consistent with with checkstyle's license and give more freedom for user, as they could choose any version they want from v2.1 to any later version.

Please confirm that you are ok with re-licensing.
Or let me know if there is problem with this. 

thanks,
Roman Ivanov
-- 
You received this message because you are subscribed to the Google Groups "SonarQube" group.
[...]

PR #73, opened 2017-02-14
Issue #45: resolved multiple simple formatting violations

rnveach:
Issue #45

Mostly just ran eclipse cs' formatter and fixed minor things resulting from any issues it made. Split this off from any of the others as they will probably be more involved.
Did apply the header to all Java files as it was a simple copy/paste.


romani:
please make special header file that is for checkstyle but for LGPL v3.
in main project we have LGPL v2. Sonar team is not ok to downgrade licensing, they did this on good reason. 

I updated #60  with reply from Sonarqube.

rnveach:
@romani I changed the standard header from 2.1 to 3. If something more is needed, please let me know.

Issue #94, opened 2017-03-28
LineLengthCheck and IndentationCheck should support property tabWidth, as it is not possible to set it for TreeWalker module


amaramrahul:
I would like to set the tabWidth property of TreeWalker module as described in http://checkstyle.sourceforge.net/config.html#Properties#TreeWalker

I don't see any option for doing this. I do not mind recompiling your module if it is necessary for adding this support.

Kindly advice.

rnveach:
@amaramrahul 
You can't specifically specify `TreeWalker` in sonarqube as a rule. Is there a specific check that you are trying to activate `tabWidth` for?
The only 2 checks that use it, by scanning Checkstyle sources, are `LineLengthCheck` and `IndentationCheck`.

@romani 
None of the modules above specify `tabWidth` as a property.
`TreeWalker` is an `AbstractFileSetCheck`. Most likely the missing `tabWidth` property is a result of mindlessly removing `AbstractFileSetCheck` properties from all classes that implement it. See https://github.com/checkstyle/sonar-checkstyle/blob/5ec7ca9fb064bf39c1e200d8f8db183c5bdeb66f/checkstyle-sonar-plugin/src/test/java/org/sonar/plugins/checkstyle/internal/ChecksTest.java#L366-L371 .

amaramrahul:
Yes, I need it for LineLengthCheck. How do I go about configuring it in sonarqube?

amaramrahul:
Would adding the tabWidth property to LineLengthCheck in checkstyle-sonar-plugin/src/main/resources/org/sonar/plugins/checkstyle/rules.xml and recompiling the plugin help? Or is there a cleaner way of achieving the same?

I would prefer to avoid recompiling the checkstyle module if possible.

romani:
http://checkstyle.sourceforge.net/config_sizes.html#LineLength

There are not property `tabWidth` in documentation, but there is in code https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/api/AbstractCheck.java#L231

It should be possible to be done from sonar UI if you update XML configs that sonar use to manage metadata.

romani:
```
$ cat MyClass.java 
public class MyClass {
	int i;
}

$ cat config.xml 
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
[...]

PR #95, opened 2017-03-28
Issue #94: added support for tabWidth in specific checks

rnveach:
Issue #94 

Added override for 2 checks.


rnveach:
@romani Tested.

It seems when new property is added to existing rule, it auto-populates default value.
When that property is taken away, it doesn't show it anymore and doesn't populate it into the generated configuration file.

Issue #93, opened 2017-03-26
upgrade to checkstyle 7.6.1


romani:
http://checkstyle.sourceforge.net/releasenotes.html#Release_7.6.1 , 

upgrade version at https://github.com/checkstyle/sonar-checkstyle/edit/master/pom.xml#L88

romani:
After upgrade user will need to reactivate "PackageDeclaration" , as previous rule will be disabled, see picture below. That is expected as Check got new property so now user should configure it before usage.
![screenshot-upgrade-tocheckstyle-7 6 1-package-declaration](https://cloud.githubusercontent.com/assets/812984/24831198/2379945c-1c4a-11e7-8523-d3708bb48889.png)

romani:
fix is merged

PR #97, opened 2017-03-28
Issue #93: upgrade to checkstyle 7.6.1

romani:
Issue #93


checkstyle-sonar-plugin/src/main/resources/org/sonar/plugins/checkstyle/rules.xml
```diff
@@ -1245,6 +1245,10 @@
     <name><![CDATA[Package Declaration]]></name>
     <configKey><![CDATA[Checker/TreeWalker/PackageDeclaration]]></configKey>
     <status>READY</status>
+    <cardinality>MULTIPLE</cardinality>
```
rnveach: cardinality shouldn't be multiple.
We only have 1 property that is either true/false for this check. Users won't activate 2 instances with both values on.

romani: I missed your point, and did a release.
lets keep it as is , I will be more accurate/attentive later on.

romani:
merged manually.

checkstyle-sonar-plugin/src/main/resources/org/sonar/plugins/checkstyle/rules.xml
rnveach: Will be fixed in cardinality UT issue.

Issue #130, opened 2017-11-07
Wrong checkstyle version output when using the plugin


pmayweg:
Sonar version: 6.4
Checkstyle sonar plugin verson: 3,7

When you switch on debug output, the checkstyle version is displayed 7,3, which is wrong

The reason is a surplus maven property in checkstyle-sonar-plugin/pom.xml

romani:
fixes are merged

PR #132, opened 2017-11-07
minor: fix ReturnCount checkstyle violations

romani:


romani:
merged manually, squashed.

fix is related to #130

Issue #145, opened 2018-03-01
Support multiple Separator Wrap rule instances


jgchristian:
Sonar version: 6.7.1
Checkstyle sonar plugin verson: 4.8

Steps to reproduce or description of problem:
- Attempt to activate multiple instances of the "Separator Wrap" rule to line up with Google Style Guide: https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml#L84
- Presently this rule is not configured in the plugin as having multiple cardinality (i.e. a Sonar rule template) so only one instance can be activated.

====================
Migration notes:
On existing rules there will be warning " This rule template was activated on the following profiles in previous versions of SonarQube. It is not possible anymore to do so. Instead, please create a custom rule. "
It will continue to work as before, to resolve warning just recreate Rule in SoanrQube.

rnveach:
This check is probably not the only one who needs to be fixed.
Eventually the UT should validate the check's carnality defined by the rules setup in https://github.com/checkstyle/sonar-checkstyle/issues/87 .

rnveach:
Fix was merged

PR #146, opened 2018-03-01
Issue #145: Templatise the Separator Wrap rule 

jgchristian:
Issue #145 change to support multiple instances. Not sure what would happen if people already have this rule configured in a singular manner.

### DIFF
--- a/checkstyle-sonar-plugin/src/main/resources/org/sonar/plugins/checkstyle/rules.xml
+++ b/checkstyle-sonar-plugin/src/main/resources/org/sonar/plugins/checkstyle/rules.xml
@@ -2028,6 +2028,7 @@
     <priority>MINOR</priority>
     <name><![CDATA[Separator Wrap]]></name>
     <configKey><![CDATA[Checker/TreeWalker/SeparatorWrap]]></configKey>
+    <cardinality>MULTIPLE</cardinality>
     <param key="option" type="s[eol,nl]">
       <defaultValue>eol</defaultValue>
     </param>


jgchristian:
Tested how my change handles the upgrade.  Seems fine.  The old singular rule continues to be applied, even though a warning is displayed

Steps to verify
- With plugin 4.8 installed via Marketplace, activate Separator Wrap rule on a profile.
- Stop Sonar
- Deploy version of plugin with the change to Sonar
- Start Sonar
- Navigate to the rule (which is now tagged as a Template)
- Warning displayed " This rule template was activated on the following profiles in previous versions of SonarQube. It is not possible anymore to do so. Instead, please create a custom rule. "
- Activate a scan on a project with the profile enabled
- Inspect the target/sonar/checkstyle.xml file downloaded from Sonar
-- It continues to include the old singular rule <module name="SeparatorWrap"><property name="severity" value="info"/><property name="tokens" value="DOT,COMMA"/><property name="option" value="eol"/></module>
- Back in Sonar, create a custom rule from the new template and assign to a profile
-- Separator Wrap page shows it as both a rule (with warning mentioned above) in the Quality Profiles sections  AND a Custom Rule
- Activate a scan on a project with the profile enabled
- Inspect the target/sonar/checkstyle.xml file downloaded from Sonar
-- see both rules active in the generated file

romani:
Thanks a lot for detailed testing steps it awesome.

Please do one update for git commit message and add reference to issue as prefix "Issue #145: ..."
We use this for release notes generation, release will be at this weekend.

Please move side effect description to issue, users will find it base on release notes quicker than in PR, it will help them to understand why it happens.

jgchristian:
@romani added commit as suggested

romani:
thanks a lot, but please amend existing commit, there should be only one commit.
thanks in advance.

agsmorodin:
Can we expect this fix to be a part of a release soon?

romani:
Sorry that we have delay in releases, you can always refer to dates of milestones at GitHub, I keep them actual always.

We had a delay with release of main library ( we are blocked by sourceforge issue, broken ssh access), plugin is attached to main library for now. 

Expect release at the last weekend of March.
If needed it sooner, package plugin yourself and substitute jar in sonar plugin folder.

PR #169, opened 2018-12-06
config: bump surefire to latest, to avoid strange try to dowload 'maven-surefire-plugin:3.0.0-M1'

romani:


rnveach:
@romani Now TC has new violations.

I only looked at one class' `Unused import` and it looks to be wrong.

romani:
Ok, at least we now see report

Issue #191, opened 2019-02-01
SonarQube 7.6 compatibility


Vampire:
While plugin version 4.17 is not even listed on https://docs.sonarqube.org/display/PLUG/Plugin+Version+Matrix, it also is not compatible with SonarQube 7.6.

If you try to do an analysis, you get

```exception
Caused by: java.lang.IllegalStateException: Unable to load component class org.sonar.plugins.checkstyle.CheckstyleConfiguration
        at org.sonar.core.platform.ComponentContainer$ExtendedDefaultPicoContainer.getComponent(ComponentContainer.java:65)
        at org.picocontainer.DefaultPicoContainer.getComponent(DefaultPicoContainer.java:632)
        at org.picocontainer.parameters.BasicComponentParameter$1.resolveInstance(BasicComponentParameter.java:118)
        at org.picocontainer.parameters.ComponentParameter$1.resolveInstance(ComponentParameter.java:136)
        at org.picocontainer.injectors.SingleMemberInjector.getParameter(SingleMemberInjector.java:78)
        at org.picocontainer.injectors.ConstructorInjector$CtorAndAdapters.getParameterArguments(ConstructorInjector.java:309)
        at org.picocontainer.injectors.ConstructorInjector$1.run(ConstructorInjector.java:335)
        at org.picocontainer.injectors.AbstractInjector$ThreadLocalCyclicDependencyGuard.observe(AbstractInjector.java:270)
        at org.picocontainer.injectors.ConstructorInjector.getComponentInstance(ConstructorInjector.java:364)
        at org.picocontainer.injectors.AbstractInjectionFactory$LifecycleAdapter.getComponentInstance(AbstractInjectionFactory.java:56)
        at org.picocontainer.behaviors.AbstractBehavior.getComponentInstance(AbstractBehavior.java:64)
        at org.picocontainer.behaviors.Stored.getComponentInstance(Stored.java:91)
        at org.picocontainer.DefaultPicoContainer.getInstance(DefaultPicoContainer.java:699)
        at org.picocontainer.DefaultPicoContainer.getComponent(DefaultPicoContainer.java:647)
        at org.sonar.core.platform.ComponentContainer$ExtendedDefaultPicoContainer.getComponent(ComponentContainer.java:63)
        ... 172 more
Caused by: org.picocontainer.injectors.AbstractInjector$UnsatisfiableDependenciesException:
[...]

PR #193, opened 2019-02-05
#191 - add support for SQ 7.6

muhlba91:
This adds SQ 7.6 support by removing the deprecated API (RulesProfile on ScannerSide). Tested manually on a SQ 7.6-community instance and analyzing https://github.com/SonarSource/sonar-scanning-examples/tree/master/sonarqube-scanner-gradle).


muhlba91:
Except for the CI builds now since I have increased SQ version to 7.6 (I'll try reverting to LTS once implementation is done), I have troubles figuring out how I got the branch coverage < 82%: `[ERROR] project failed coverage check. Branch coverage rate of 81.9% is below 82.0%`. Any ideas?

romani:
>Any ideas?

Please raise coverage in some other place. We are not very strict in this project, just try to not do worse. Coverage still matter as we have to page gate at sonarqube to be published to their plugin center

muhlba91:
@romani - please review. :)
It would be great to also have a SQ 7.6 build image for the CI to make sure compatibility is given to 6.7 and 7.6 per CI testing. I have increased the sleep time to 60 seconds due to hitting the 40 seconds startup limit quite a few times now.

I have tried keeping the minimum version down at 6.7 which works because I have mocked the `ActiveRules` in the tests return a custom class implementing `ActiveRule` when queried for rules. The issue here is that SQ changed the way they create their internal class `NewActiveRule` in SQ 7.5 - from constructor to builder pattern and, in fact, makes it incompatible. Hence, I implemented a test class (https://github.com/checkstyle/sonar-checkstyle/pull/193/files#diff-dd61f11ee0e98cec2392a4b598d09f27R264) to keep compatibility down, once compatibility is raised to at least 7.5 we can think of using the `NewActiveRule` builder for the test.

Otherwise, I have tested it on 7.6-community locally (fresh server installation) using above mentioned project for analysis.

However, the annotation`@ScannerSide` is deprecated (https://github.com/SonarSource/sonarqube/blob/master/sonar-plugin-api/src/main/java/org/sonar/api/batch/ScannerSide.java) and should be replaced in the future. However, I tried replacing it but ran into issues with the `JavaResourceLocator` not being instantiated and injected anymore. I guess SQ has restructured some parts and once they remove the annotation and injection still doesn't work by then, I guess a restructuring of corresponding classes has to be done. From what I've got now (please correct me) a ProfileExporter should not run on the scanner side, for example. Nevertheless, further tests and investigation can be done once issues arise. ;)

romani:
@muhlba91 , please use docker image for 7.6 - https://hub.docker.com/r/checkstyle/sonarqube-maven-git/tags, I just uploaded it.

checkstyle-sonar-plugin/src/test/java/org/sonar/plugins/checkstyle/CheckstyleSeverityUtilsTest.java
```diff
@@ -37,6 +37,20 @@ public void testToSeverity() {
...
+    @Test
+    public void testToSeverityString() {
+        assertThat(CheckstyleSeverityUtils.toSeverity("BLOCKER")).isEqualTo("error");
+        assertThat(CheckstyleSeverityUtils.toSeverity("CRITICAL")).isEqualTo("error");
+        assertThat(CheckstyleSeverityUtils.toSeverity("MAJOR")).isEqualTo("warning");
+        assertThat(CheckstyleSeverityUtils.toSeverity("MINOR")).isEqualTo("info");
+        assertThat(CheckstyleSeverityUtils.toSeverity("INFO")).isEqualTo("info");
+    }
+
+    @Test(expected = IllegalArgumentException.class)
```
romani: convert to catch.

checkstyle-sonar-plugin/src/test/resources/org/sonar/plugins/checkstyle/CheckstyleProfileExporterTest/singleCheckstyleActiveRulesToExport.xml
```diff
@@ -0,0 +1,12 @@
...
+    <module name="SuppressWarningsFilter"/>
+    <module name="JavadocPackage">
+        <property name="id" value="com.puppycrawl.tools.checkstyle.checks.javadoc.JavadocPackageCheck"/>
+        <property name="severity" value="warning"/>
+    </module>
+    <module name="TreeWalker">
+        <module name="SuppressWarningsHolder"/>
+        <module name="SuppressionCommentFilter"/>
+    </module>
+</module>
```
romani: add \n

checkstyle-sonar-plugin/src/test/java/org/sonar/plugins/checkstyle/CheckstyleProfileExporterTest.java
```diff
@@ -109,6 +120,33 @@ public void singleCheckstyleRulesToExport() {
                         + "singleCheckstyleRulesToExport.xml", sanitizeForTests(writer.toString()));
     }
 
+    @Test(expected = IllegalStateException.class)
```
romani: do it by `catch`.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleProfileExporter.java
```diff
@@ -63,19 +67,39 @@ public void exportProfile(RulesProfile profile, Writer writer) {
...
+                        arrangeByConfigKey(activeRuleWrappers);
                 generateXml(writer, activeRulesByConfigKey);
             }
         }
         catch (IOException ex) {
             throw new IllegalStateException("Fail to export the profile " + profile, ex);
         }
+    }
+
+    protected void exportProfile(ActiveRules activeRules, Writer writer) {
```
romani: looks like you do `protected` for UT, it is not good.
I know there bunch code around that do the same, it is very bad practice, that is left from previous maintainers.

UT should be able to pass all branches through public entry points, in case where need to cover exceptional cases it is better to use reflection and explain in comment over test why reflection is the only way out.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleConfiguration.java
```diff
@@ -54,18 +54,18 @@
...
-    private final RulesProfile profile;
+    private final ActiveRules activeRules;
     private final org.sonar.api.config.Configuration conf;
     private final FileSystem fileSystem;
 
     public CheckstyleConfiguration(
             org.sonar.api.config.Configuration conf,
             CheckstyleProfileExporter confExporter,
-            RulesProfile profile,
+            ActiveRules activeRules,
```
romani: @muhlba91 , please help me to understand a problem.
I still see this class in api of sonarqube, I placed link to issue description, so class is not deprecated, content of it heavily deprecated. Why do you think we change c-tor type ?

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/rule/ActiveRuleWrapperBatchImpl.java
```diff
@@ -0,0 +1,58 @@
...
+////////////////////////////////////////////////////////////////////////////////
+
+package org.sonar.plugins.checkstyle.rule;
+
+import java.util.Map;
+
+import org.sonar.api.batch.rule.ActiveRule;
+import org.sonar.plugins.checkstyle.CheckstyleSeverityUtils;
+
+public class ActiveRuleWrapperBatchImpl implements ActiveRuleWrapper {
```
romani: we need to add javadoc and explain that we need such classes only till we upgrade to 7.X LST.
and some details as you did in PR comment.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleConfiguration.java
muhlba91: In short, SQ has removed the access to this API only for ScannerSide (https://docs.sonarqube.org/plugins/servlet/mobile?contentId=1442007#content/view/1442007). This is why it's still working for the exporter on the server side. As mentioned in the comment already, we are kinda abusing the exporter for generating the checkstyle file on scanner side, because an exporter shall not be available/instantiated on scanner side and, with deprecation of the old `ScannerSide` side annotation and introducing the new one for 7.6, we will get a problem once they remove the old one because then no `JavaResourceLocator` will be available on ScannerSide on which our exporter relies on. In fact, we would need to refactor the exporter to be only on server side and create a new class for writing the configuration on the scanner side (e.g. through helper methods/classes).
This is why we have to replace `RulesProfile` here as we are working on ScannerSide where it's not available anymore.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/rule/ActiveRuleWrapperBatchImpl.java
muhlba91: I added a JavaDoc comment for the interface and here as a short reference to what the wrapper is wrapping.
However, as noted in the other JavaDoc, for now we have to use the wrapper unless we refactor writing the checkstyle configuration from the exporter which needs `RulesProfile` and we don't want to copy paste the entire class only for the new `ActiveRules` scanner side part.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleProfileExporter.java
muhlba91: No, it's `protected` because of `https://github.com/checkstyle/sonar-checkstyle/pull/193/files#diff-515b093b908aa69201af7db1921c8263R77`. I could make it public as well just like the implemented abstract method but wanted to keep visibility as low as possible as it is kinda a "convenience method" imo to make use of the same exporter on the scanner side.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleConfiguration.java
romani: > Should we create an issue for that?

sure, please be welcome.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleProfileExporter.java
romani: please make it public.
We already have public method `exportProfile` with different arguments, lets have one more, especially if it suppose to be used by another class.

protected give only some strange illusion of restricted access.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/rule/ActiveRuleWrapperBatchImpl.java
```diff
@@ -0,0 +1,61 @@
...
+
+package org.sonar.plugins.checkstyle.rule;
+
+import java.util.Map;
+
+import org.sonar.api.batch.rule.ActiveRule;
+import org.sonar.plugins.checkstyle.CheckstyleSeverityUtils;
+
+/**
+ * Wrapper as per {@link ActiveRuleWrapper} for the new scanner side {@link ActiveRule}.
```
romani: `Implementation of {@link ActiveRuleWrapper} for the new scanner side {@link org.sonar.api.batch.rule.ActiveRule}`

full name is used to make it more clear, as too much the same name classes.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/rule/ActiveRuleWrapperImpl.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.Map;
+import java.util.Objects;
+
+import org.apache.commons.lang.StringUtils;
+import org.sonar.api.rules.ActiveRule;
+import org.sonar.api.rules.RuleParam;
+import org.sonar.plugins.checkstyle.CheckstyleSeverityUtils;
+
+/**
+ * Wrapper as per {@link ActiveRuleWrapper} for the server side {@link ActiveRule}.
```
romani: `Implementation of {@link ActiveRuleWrapper} for the new server side {@link org.sonar.api.rules.ActiveRule}`

full name is used to make it more clear, as too much the same name classes.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/rule/ActiveRuleWrapperBatchImpl.java
romani: please  create issue on this.

romani:
@muhlba91 , looks like now you are the most experienced engineer in this project, let me know if you interested to become a maintainer of it.

muhlba91:
I have created issue #194 regarding `ScannerSide` annotations and fixed the requested changes - how can I get rid of TC errors now since those full imports in the JavaDoc are not necessary?

@romani thank you! Sure, I'd be delighted being able to contribute more and be interested in becoming a maintainer. :)

tsjensen:
@muhlba91 Thank you for all your work on sonar-checkstyle. Highly appreciated! 👍

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/rule/ActiveRuleWrapperBatchImpl.java
```diff
@@ -0,0 +1,62 @@
...
+package org.sonar.plugins.checkstyle.rule;
+
+import java.util.Map;
+
+import org.sonar.api.batch.rule.ActiveRule;
+import org.sonar.plugins.checkstyle.CheckstyleSeverityUtils;
+
+/**
+ * Wrapper as per {@link ActiveRuleWrapper} for the
+ * new scanner side {@link org.sonar.api.batch.rule.ActiveRule}.
```
rnveach: Swap `org.sonar.api.batch.rule.ActiveRule` with `ActiveRule`.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/rule/ActiveRuleWrapperImpl.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Objects;
+
+import org.apache.commons.lang.StringUtils;
+import org.sonar.api.rules.ActiveRule;
+import org.sonar.api.rules.RuleParam;
+import org.sonar.plugins.checkstyle.CheckstyleSeverityUtils;
+
+/**
+ * Wrapper as per {@link ActiveRuleWrapper} for the
+ * server side {@link org.sonar.api.rules.ActiveRule}.
```
rnveach: Swap `org.sonar.api.rules.ActiveRule` with `ActiveRule`.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/rule/ActiveRuleWrapperBatchImpl.java
romani: @muhlba91, Ok, let's keep simple name.

romani:
@muhlba91 , please recheck invite in email, for now only read access, but will give me ability to assign you to issues if you place in comment "I am on it.", ones you made more than 15 commits, you can get RW access.
But no matter what level permission you have, you have to pass review from two admins (me and @rnveach). Only admins can pass only one review from another admin. It slow down development a bit, but proven numerous of times that it save a lot cases of regression and prevents incomplete/bad decisions.

Welcome to the team, feel free to grab any issue (create new) and improve all that you want. 
FYI, we (admins) are mostly busy in core library project.

muhlba91:
@rnveach @romani please review and merge. :)

Thank you! Sure, also from my experience it's always best to have on average if possible more than 1 person reviewing to catch possible regressions.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleProfileExporter.java
```diff
@@ -63,19 +67,39 @@ public void exportProfile(RulesProfile profile, Writer writer) {
...
         }
         catch (IOException ex) {
             throw new IllegalStateException("Fail to export the profile " + profile, ex);
         }
+    }
+
+    public void exportProfile(ActiveRules activeRules, Writer writer) {
+        try {
+            final List<ActiveRuleWrapper> activeRuleWrappers = new ArrayList<>();
+            for (org.sonar.api.batch.rule.ActiveRule activeRule : activeRules
```
rnveach: This class looks imported, so you don't need the full classpath.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleProfileExporter.java
```diff
@@ -63,19 +67,39 @@ public void exportProfile(RulesProfile profile, Writer writer) {
...
                     .getActiveRulesByRepository(CheckstyleConstants.REPOSITORY_KEY);
             if (activeRules != null) {
-                final Map<String, List<ActiveRule>> activeRulesByConfigKey =
-                        arrangeByConfigKey(activeRules);
+                final List<ActiveRuleWrapper> activeRuleWrappers = new ArrayList<>();
+                for (ActiveRule activeRule : activeRules) {
+                    activeRuleWrappers.add(new ActiveRuleWrapperImpl(activeRule));
+                }
+                final Map<String, List<ActiveRuleWrapper>> activeRulesByConfigKey =
+                        arrangeByConfigKey(activeRuleWrappers);
```
rnveach: Whenever you call `arrangeByConfigKey`, right before it is an array converting a list of ActiveRules into a ActiveRuleWrapper. Since both calls do the same thing this looks like it can be placed inside the `arrangeByConfigKey` method.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleProfileExporter.java
```diff
@@ -63,19 +67,39 @@ public void exportProfile(RulesProfile profile, Writer writer) {
...
+
+    public void exportProfile(ActiveRules activeRules, Writer writer) {
+        try {
+            final List<ActiveRuleWrapper> activeRuleWrappers = new ArrayList<>();
+            for (org.sonar.api.batch.rule.ActiveRule activeRule : activeRules
+                    .findByRepository(CheckstyleConstants.REPOSITORY_KEY)) {
+                activeRuleWrappers.add(new ActiveRuleWrapperBatchImpl(activeRule));
+            }
+            final Map<String, List<ActiveRuleWrapper>> activeRulesByConfigKey =
+                    arrangeByConfigKey(activeRuleWrappers);
```
rnveach: same.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleSeverityUtils.java
```diff
@@ -34,18 +34,22 @@ private CheckstyleSeverityUtils() {
...
+        return toSeverity(priority.name());
+    }
+
+    public static String toSeverity(String priority) {
         final String result;
 
         switch (priority) {
-            case BLOCKER:
-            case CRITICAL:
+            case "BLOCKER":
```
rnveach: There is no way to convert String priority into `RulePriority` to leave the enumeration switch?

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/rule/ActiveRuleWrapperBatchImpl.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.util.Map;
+
+import org.sonar.api.batch.rule.ActiveRule;
+import org.sonar.plugins.checkstyle.CheckstyleSeverityUtils;
+
+/**
+ * Wrapper as per {@link ActiveRuleWrapper} for the new scanner side {@link ActiveRule}.
+ */
+public class ActiveRuleWrapperBatchImpl implements ActiveRuleWrapper {
```
rnveach: Would it make more sense to name this `...ScannerImpl`?

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/rule/ActiveRuleWrapperImpl.java
```diff
@@ -0,0 +1,76 @@
...
+
+import org.apache.commons.lang.StringUtils;
+import org.sonar.api.rules.ActiveRule;
+import org.sonar.api.rules.RuleParam;
+import org.sonar.plugins.checkstyle.CheckstyleSeverityUtils;
+
+/**
+ * Wrapper as per {@link ActiveRuleWrapper} for the server side {@link ActiveRule}.
+ */
+public class ActiveRuleWrapperImpl implements ActiveRuleWrapper {
```
rnveach: Would it make more sense to name this `...ServerImpl`?

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleProfileExporter.java
muhlba91: No, unfortunately not. That's the issue I explained in a comment earlier that the `ActiveRule` from `RulesProfile` is server sided and a new `ActiveRule` (different package) from `ActiveRules` is scanner sided. Therefore, I convert the corresponding `ActiveRule` to the wrapper which returns the values we need in the exporter.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleSeverityUtils.java
muhlba91: Well, there's another issue with `RulePriority` -> it's deprecated since 4.2. The new `ActiveRule` implementation only returns the String value. I actually don't know if the annotations of SQ make sense because they deprecate `RulePriority` but don't deprecate `ActiveRule::getSeverity` (server side) which in turn returns `RulePriority` but they deprecated the method `getPriority`.
Hence, I see two options: we convert the String from the new `ActiveRules` to the deprecated `RulePriority` and leave the enumeration switch or we use a String switch but, personally, I'd prefer not using deprecated classes?

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleProfileExporter.java
muhlba91: No, this is the new canner side `ActiveRule`, only the server side `ActiveRule` is imported.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleSeverityUtils.java
rnveach: Ok, ignore this for now.
Is there any chance to remove this deprecated usage then?

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleProfileExporter.java
rnveach: My bad, import had an `s` on the end. It was `ActiveRules` being imported.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleProfileExporter.java
rnveach: Sorry, I missed the different class names in the `new`.

rnveach:
@romani Please finalize.

romani:
@muhlba91 ,

> **muhlba91** marked this conversation as resolved.

it would be good if you also type  "done" to make sure that you addressed item, usually "resolve conversation" is done by author of review point. To make sure he is agree with fix.

PR #200, opened 2019-03-21
Fix auto unboxing issue in CheckstyleAuditListener

### LINKED ISSUE #199: Auto-unboxing issue in CheckstyleAuditListener
Sonar version: 7.6
Checkstyle sonar plugin verson: 4.18

Steps to reproduce or description of problem:

Run any check which reports line 0 for an issue. This used to be the required behavior for checks that flag the entire file, and have no relation to the AST. For example FileSetChecks.  
I know there was a breaking change in checkstyle/checkstyle#6000, changing the "magic" line number to 1. This kind of neutralizes this problem.  
BUT: I noticed the problem running custom checks at work. I would expect some custom checks to be around that haven't noticed checkstyle/checkstyle#6000.

The error caused is this:
```
Caused by: com.puppycrawl.tools.checkstyle.api.CheckstyleException: Exception was thrown while processing C:\path\to\File.java
        at com.puppycrawl.tools.checkstyle.Checker.processFiles(Checker.java:304)
        at com.puppycrawl.tools.checkstyle.Checker.process(Checker.java:217)
        at org.sonar.plugins.checkstyle.CheckstyleExecutor.executeWithClassLoader(CheckstyleExecutor.java:111)
        ... 71 more
Caused by: java.lang.NullPointerException
        at org.sonar.plugins.checkstyle.CheckstyleAuditListener.addError(CheckstyleAuditListener.java:106)
        at com.puppycrawl.tools.checkstyle.Checker.fireErrors(Checker.java:401)
        at com.puppycrawl.tools.checkstyle.Checker.processFiles(Checker.java:292)
        ... 73 more
```

I have already fixed this, it's a small thing. PR upcoming.

tsjensen:
Fixes #199.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ Icon?
 # Files that might appear on external disk
 .Spotlight-V100
 .Trashes
+_support/
 
 # ---- Windows
 # Windows image file caches
--- a/checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleAuditListener.java
+++ b/checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleAuditListener.java
@@ -100,15 +100,10 @@ public class CheckstyleAuditListener implements AuditListener {
             final NewIssue issue = context.newIssue();
             final ActiveRule rule = ruleFinder.find(
                     RuleKey.of(CheckstyleConstants.REPOSITORY_KEY, ruleKey));
-            final Integer lineId = getLineId(event);
-            int lineNo = 1;
-            if (lineId != null) {
-                lineNo = lineId;
-            }
             if (Objects.nonNull(issue) && Objects.nonNull(rule)) {
                 final NewIssueLocation location = issue.newLocation()
                         .on(currentResource)
-                        .at(currentResource.selectLine(lineNo))
+                        .at(currentResource.selectLine(getLineId(event)))
                         .message(message);
                 issue.forRule(rule.ruleKey())
                         .at(location)
@@ -162,14 +157,12 @@ public class CheckstyleAuditListener implements AuditListener {
     }
 
     @VisibleForTesting
-    static Integer getLineId(AuditEvent event) {
-        Integer result = null;
+    static int getLineId(AuditEvent event) {
+        int result = 1;
         try {
-            final int line = event.getLine();
-            // checkstyle returns 0 if there is no relation to a file content,
-            // but we use null
-            if (line != 0) {
-                result = line;
+            final int eventLine = event.getLine();
+            if (eventLine > 0) {
+                result = eventLine;
             }
         }
         catch (Exception ex) {
--- a/checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleAuditListener.java
+++ b/checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleAuditListener.java
@@ -100,10 +100,15 @@ public class CheckstyleAuditListener implements AuditListener {
             final NewIssue issue = context.newIssue();
             final ActiveRule rule = ruleFinder.find(
                     RuleKey.of(CheckstyleConstants.REPOSITORY_KEY, ruleKey));
+            final Integer lineId = getLineId(event);
+            int lineNo = 1;
+            if (lineId != null) {
+                lineNo = lineId;
+            }
             if (Objects.nonNull(issue) && Objects.nonNull(rule)) {
                 final NewIssueLocation location = issue.newLocation()
                         .on(currentResource)
-                        .at(currentResource.selectLine(getLineId(event)))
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -20,6 +20,7 @@ Icon?
 # Files that might appear on external disk
 .Spotlight-V100
 .Trashes
+_support/
```
rnveach: Just curious what creates this file in the workspace?

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleAuditListener.java
```diff
@@ -100,10 +100,11 @@ public void addError(AuditEvent event) {
             final NewIssue issue = context.newIssue();
             final ActiveRule rule = ruleFinder.find(
                     RuleKey.of(CheckstyleConstants.REPOSITORY_KEY, ruleKey));
+            final Integer lineId = getLineId(event);
             if (Objects.nonNull(issue) && Objects.nonNull(rule)) {
                 final NewIssueLocation location = issue.newLocation()
                         .on(currentResource)
-                        .at(currentResource.selectLine(getLineId(event)))
+                        .at(currentResource.selectLine(lineId != null ? lineId.intValue() : 1))
```
rnveach: `getLineId` is only used in this class and method by us besides tests. Why do we want the method to keep returning `null` when no one will use the `null` value. Shouldn't we just change it's default to `1` and have it return `int` instead?

tsjensen: True, I chose the least invasive change that fixes the problem. After it's fixed, I agree that it may be a good idea to refactor `getLineId()` as well.  
But I must admit that I don't know enough about the underlying mechanics of SonarQube - maybe they really do use `null` to signify that an issue relates to the entire file. Then by changing these semantics for us, we might break conceptual integrity. I didn't want to take any risks here. 😉

.gitignore
tsjensen: It's just a local tool in our toolchain, which generates some files that we don't want committed.
If it's a problem to have this entry, we can leave it out. I mostly added it for my own convenience.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleAuditListener.java
rnveach: > maybe they really do use null to signify that an issue relates to the entire file

The interface for InputFile says it can't be `null` by its definition,
> `TextRange selectLine(int line);`

This is what is causing the auto-unboxing. If it is true they allow `null` the only way to distinguish this is to call `NewIssueLocation#at` with the entire range of the file. As you mentioned this is related to https://github.com/checkstyle/checkstyle/issues/6000 , which stated it was better to print violation on some physical line in file, hence why we picked line 1. So we are no longer printing violations on entire file.

So I see no issues making the change.

tsjensen: (Well, line 1 might be problematic if the file is empty. This will have to be handled in IDE plugins, which must deal with the fact that issues are reported on non-existent lines. Not a big deal, but as we see, it should be handled consistently throughout our little Checkstyle ecosystem. But that's really a Checkstyle issue rather than one here.)
Do you need *me* to also do the refactoring of `getLineId()`? Or would it be acceptable to have this fix first, and do the refactoring later?

rnveach: > Well, line 1 might be problematic if the file is empty.

Then this is also an issue with your change, since you are doing the same thing I am asking for just in a different place. You are replacing `null` with `1` without checking the size of the file. We changed all 0s to 1s and didn't do any change here to check number of lines in file being reported, so we don't even know if the 1 reported by checkstyle is valid.
So this doesn't help your case to make the change later. :)

This raises another question outside of this issue, what happens if we count lines differently then sonar (mis-configuration) and we print a violation on a line which it thinks is outside it's understanding of the size of the file.

> Or would it be acceptable to have this fix first, and do the refactoring later?

I prefer the fix now as I don't see any difference in hurt between your version and what I am asking it to be.

rnveach:
@tsjensen is it also possible to add a test to sonar to show this issue so we don't have future regression?

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleAuditListener.java
tsjensen: ok - stand by ...

romani:
@muhlba91 , please help with review.

tsjensen:
Done.

rnveach:
Thanks.

rnveach:
@romani Please be the final review.

Issue #208, opened 2019-04-12
Sonarqube 7.7 support (sonar-java-plugin-5.12)


charlesfendt:
Sonar version: 7.7
Checkstyle sonar plugin verson: 4.18

Steps to reproduce or description of problem:
- Update to Sonarqube 7.7 and checkstyle plugin 4.18
- try to compile any project

Caused by: java.lang.NoSuchMethodError: org.sonar.plugins.java.api.JavaResourceLocator.findSourceFileKeyByClassName(Ljava/lang/String;)Ljava/lang/String;
    at org.sonar.plugins.findbugs.resource.ByteCodeResourceLocator.findSourceFileKeyByClassName (ByteCodeResourceLocator.java:59)

[sonar-7.7.log](https://github.com/checkstyle/sonar-checkstyle/files/3074479/sonar-7.7.log)

charlesfendt:
I got the same error on SQ 7.6 if I perform the plugin update...
sonar-java-plugin-5.12.0.17701.jar seems to be the root cause of the error...
with sonar-java-plugin-5.10.1.16922.jar, all is working fine :-/

tsjensen:
Interesting.  
At least for me, I can say that 7.7+4.18 worked. It also worked for @marcofranssen ([link](https://github.com/checkstyle/sonar-checkstyle/issues/205#issuecomment-478999706)).

charlesfendt:
I perform the upgrade of SQ yesterday to version 7.7 -> analyze failure (see attached log). I tried an upgrade de 7.6 (my VM snapshot is in 7.2.1). I make always all plugins updates... same crash!
I tried one last time with SQ 7.6,all non-“Sonar-Java” plugins update and it works!

What is your java plugin version?

romani:
https://github.com/checkstyle/sonar-checkstyle/blob/master/pom.xml#L89

It is 5.9.1.16423

charlesfendt:
5.12.0.17701 must be the problem :-(

romani:
We need to get rid of this dependency at all , see #40

muhlba91:
[...]

PR #209, opened 2019-04-15
Update to CS 8.19 and gain SQ 7.7+ compatibility

### LINKED ISSUE #204: upgrade to checkstyle 8.19
https://checkstyle.org/releasenotes.html#Release_8.19 , upgrade version at https://github.com/checkstyle/sonar-checkstyle/edit/master/pom.xml#L88

Migration notes for users:
1) AnnotationLocation got [two new default tokens](https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/checks/annotation/AnnotationLocationCheck.java#L245-L256). Existing Sonar Rules based on AnnotationLocation Check will continue to use old set of tokens. To use new tokens, you need simply recreate Rule for AnnotationLocation.

### LINKED ISSUE #206: Introduce tabWidth to Checker
As per #204 and the corresponding changeset in CS (https://github.com/checkstyle/checkstyle/issues/6485), we have to add the property `tabWidth` to all checks and update the translations as well.

muhlba91:
* fixes #204 
* fixes #206 
* update sonar-java as per #208


muhlba91:
please rerun wercker build

rnveach:
I restarted wercker.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleProfileExporter.java
```diff
@@ -118,6 +119,12 @@ private void appendCustomFilters(Writer writer) throws IOException {
         writer.append(filtersXml);
     }
 
+    private void appendTabWidth(Writer writer) throws IOException {
+        final String tabWidth = configuration.get(CheckstyleConstants.CHECKER_TAB_WIDTH)
+                .orElse(null);
```
rnveach: I don't think we should default to `null`.  Use the new constant in `CheckstyleConstants`.

rnveach: Nevermind, if it is `null` the property isn't written to the writer. So it will use the default of checkstyle.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstylePlugin.java
```diff
@@ -77,6 +84,15 @@ public List getExtensions() {
                                 .description(TREEWALKER_FILTERS_DESCRIPTION)
                                 .type(PropertyType.TEXT)
                                 .onQualifiers(Qualifiers.PROJECT, Qualifiers.MODULE).build(),
+                        PropertyDefinition.builder(CheckstyleConstants.CHECKER_TAB_WIDTH)
+                                .defaultValue(CheckstyleConstants.CHECKER_TAB_WIDTH_DEFAULT_VALUE)
```
rnveach: Should we not specify a default and allow it to be a blank string?
This way we use main repo's default and not overwrite our own. Description should be updated to say what a blank string means if we do this.

muhlba91: Makes sense, I have left it empty by default and added a note in the description.

rnveach:
@romani Will you be reviewing this?

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleConstants.java
```diff
@@ -28,6 +28,7 @@
 
     public static final String CHECKER_FILTERS_KEY = "sonar.checkstyle.filters";
     public static final String TREEWALKER_FILTERS_KEY = "sonar.checkstyle.treewalkerfilters";
+    public static final String CHECKER_TAB_WIDTH = "sonar.checkstyle.tabWidth";
```
romani: Add "checker" after checkstyle.
It is better to be precise.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstylePlugin.java
```diff
@@ -57,6 +57,14 @@
             + "SuppressionCommentFilter</a>: "
             + FILTERS_DESCRIPTION_FOOTER;
 
+    private static final String CHECKER_TAB_WIDTH_DESCRIPTION = FILTERS_DESCRIPTION_HEADER
```
romani: >FILTERS_DESCRIPTION_HEADER

Please rename this variable to smth more generic

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstylePlugin.java
```diff
@@ -57,6 +57,14 @@
...
+            + " the <a href=\"http://checkstyle.sourceforge.net/config.html#tabWidth\">"
```
romani: Please use checkstyle.org in all new code.

Old places need to updated to use new url in separate issue

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleProfileExporter.java
```diff
@@ -102,6 +102,7 @@ private void generateXml(Writer writer, Map<String,
             List<ActiveRuleWrapper>> activeRulesByConfigKey) throws IOException {
         appendXmlHeader(writer);
         appendCustomFilters(writer);
+        appendTabWidth(writer);
```
romani: Please print before filters.

Config is exportable, it is better to print in good way

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleProfileExporter.java
romani: It is better to avoid null passing around, can we skip printing property instead ?

rnveach: > can we skip printing property instead ?

This is essentially what `null` does.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleConstants.java
muhlba91: Sure, I have changed it now.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstylePlugin.java
muhlba91: changed as well.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstylePlugin.java
muhlba91: Changed. I have not updated the other URLs in those decriptions as there's a separate issue for it.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleProfileExporter.java
muhlba91: moved it up to get it printed before filters.

muhlba91:
I have not received any exception when updating to the new version. What I did was creating a Sonar profile containing all CS rules and analyzed my project. After updating to the new version, I have analyzed it again and analysis ran through smoothly.
IndentationCheck and LineLengthCheck don't fail since `tabWidth` is still existing on CS side.

Can you please restart the wercker build?

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleProfileExporter.java
romani: yes, but we avoid  null usage in plugin.
So plugin will not depend on this to the library.

romani:
> I have analyzed it again and analysis ran through smoothly

good, but this mean that your profile continue to use old set of tokens. But new tokens are default tokens, so behavior in maven plugin and Sonar will be different. new tokens were added to Defaults - https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/checks/annotation/AnnotationLocationCheck.java#L245-L256
I added migration note to issue #204.

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleProfileExporter.java
muhlba91: this line filters out the `null` or empty values:
https://github.com/checkstyle/sonar-checkstyle/blob/a0330ce67100f6d11b778ed5b53afc43c86e0176/checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleProfileExporter.java#L228
I agree with @rnveach. Atm, I don't see a point in duplicating a null or empty check kinda one line before it's done anyways.

checkstyle-sonar-plugin/src/main/resources/org/sonar/l10n/checkstyle.properties
```diff
@@ -177,7 +177,6 @@ rule.checkstyle.com.puppycrawl.tools.checkstyle.checks.javadoc.JavadocStyleCheck
 rule.checkstyle.com.puppycrawl.tools.checkstyle.checks.javadoc.JavadocStyleCheck.param.checkFirstSentence=Whether to check the first sentence for proper end of sentence. Default is true.
 rule.checkstyle.com.puppycrawl.tools.checkstyle.checks.javadoc.JavadocStyleCheck.param.endOfSentenceFormat=Format for matching the end of a sentence.
 rule.checkstyle.com.puppycrawl.tools.checkstyle.checks.sizes.LineLengthCheck.name=Line Length
-rule.checkstyle.com.puppycrawl.tools.checkstyle.checks.sizes.LineLengthCheck.param.tabWidth=number of expanded spaces for a tab character (''\t'')
```
romani: please share a screenshot of how `tabWidth` is looks like in UI for this Check, I think we are loosing documentation content by removal of this line

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstylePlugin.java
romani: I just created #210

checkstyle-sonar-plugin/src/main/java/org/sonar/plugins/checkstyle/CheckstyleProfileExporter.java
romani: ok, not a critical item

checkstyle-sonar-plugin/src/main/resources/org/sonar/l10n/checkstyle.properties
muhlba91: configuration of this check:
linelength_config

rule violation in project:
linelength_project

romani: hmm, so it becomes not configurable , but it was.
Can we keep it? It should not hurt. User can override Checker's tabWidth, most users will not, but it give some flexibility to user to have few instances of Checks with different tabWidth and config and reconcile overlap/extra violations by suppressions.

Please share a screenshot of tabWidth in Checker,  my original question was to make sure that tabWidth in Checker have some hint/note like `number of expanded spaces for a tab character (''\t'')`

tsjensen: Just saw this. From a user's perspective, I should hope that my tabs have equal width in the eyes of all checks. Adding the possibility of having different tab widths per check adds a source of errors. (Per *file* might make some sense, but per *check* is confusing.)

muhlba91: I agree with @tsjensen and @rnveach proposed the deletion of tabWidth in those checks as well (https://github.com/checkstyle/sonar-checkstyle/issues/206#issuecomment-482656114).
Does the user really need to set a different tabWidth for those checks and what use cases does that cover where tabWidths are intentionally different?

No, I just referred to the documentation of this property in CS (per link). I‘ll add this sentence for clarification to the user - PR update comes tomorrow. :)

rnveach: @romani 
> We do not add tabWidth to all other Checks (for now), but lets keep smth that users requested already.

Based on your conversations before, I was going to start a breaking issue in main repo and remove tabWidth setter from everything except Checker so it is no longer a global property. This would align with how this PR is currently handling tabWidth. The reasoning was if we were against tabWidth in every check here, why would we allow main repo to do it.

I will wait until I see how we resolve this thread.

muhlba91: I added the note to the description now. Please advice which direction regarding `tabWidth` we'll be going.

romani: Please leave such properties in place in this release. Right now it is possible in library to define checker  and checks property to be different. If in future it might change, but we do know when, so we will adjust plugin accordingly.

Let's move this discussion to library project, it is not easy issue.
Plugin release should not be blocked by this.

FYI, sonarqube plugin does not critically depends on correct column reporting. Such precise position of violation is more in demand in IDE plugins, where user is positioned to exact column for fix, or used to make a fix automatically.

romani:
Please  fix second commit message.

I am not sure why I see old changes in first commit, please review. Let's get ready for merge

checkstyle-sonar-plugin/src/main/resources/org/sonar/l10n/checkstyle.properties
muhlba91: OK, reverted.

muhlba91:
@romani changed commits.

romani:
@muhlba91 , thanks a lot for update, please fix first commit message it is `@muhlba91
"#206 introduce tabWidth to Checker"` extra `"`.
It would be awesome is if follow general rule of commit messages - https://checkstyle.org/contributing.html#Submitting_your_contribution or https://github.com/checkstyle/checkstyle/blob/master/.github/PULL_REQUEST_TEMPLATE.md

So them should be:
Issue #206: introduce tabWidth to Checker
Issue #208: update sonar-java to ensure compatibility

muhlba91:
@romani - please re-run wercker. changed commit message.

romani:
@muhlba91 , please recheck all commits, I see only one as fixed.

muhlba91:
thx, i forgot to push after the second change. :)

romani:
thanks a lot, I will make release as I get to laptop

PR #214, opened 2019-04-29
Issue #187: remove checkstyle-all project

muhlba91:
fix #187 by removing `checkstyle-all`, moving all files out of `checkstyle-sonar-plugin` and merging `pom.xml`.

PR #216, opened 2019-04-30
Issue #213: update to CS 8.20

### LINKED ISSUE #213: upgrade to checkstyle 8.20
https://checkstyle.org/releasenotes.html#Release_8.20 , upgrade version at https://github.com/checkstyle/sonar-checkstyle/edit/master/pom.xml#L88

muhlba91:
Fixes #213.

Please adapt CI first after the merge of #214 to have a clean CI run for this PR.

### DIFF
--- a/src/main/resources/com/sonar/sqale/checkstyle-model.xml
+++ b/src/main/resources/com/sonar/sqale/checkstyle-model.xml
@@ -896,6 +896,19 @@
       </chc>
       <chc>
         <rule-repo>checkstyle</rule-repo>
+        <rule-key>com.puppycrawl.tools.checkstyle.checks.javadoc.MissingJavadocTypeCheck</rule-key>
+        <prop>
+          <key>remediationFunction</key>
+          <txt>CONSTANT_ISSUE</txt>
+        </prop>
+        <prop>
+          <key>offset</key>
+          <val>10</val>
+          <txt>min</txt>
+        </prop>
+      </chc>
+      <chc>
+        <rule-repo>checkstyle</rule-repo>
         <rule-key>com.puppycrawl.tools.checkstyle.checks.design.FinalClassCheck</rule-key>
         <prop>
           <key>remediationFunction</key>
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ Compatibility matrix from checkstyle team:
 
 Checkstyle Plugin|Sonar min|Sonar max|Checkstyle|Jdk
 -----------------|---------|---------|----------|---
+4.20|6.7  |7.7+|8.20|1.8
 4.19|6.7  |7.7+|8.19|1.8
 4.18|6.7  |7.7+|8.18|1.8
 4.17|6.7  |7.5|8.17|1.8
--- a/pom.xml
+++ b/pom.xml
@@ -81,7 +81,7 @@
   </ciManagement>
 
   <properties>
-    <checkstyle.version>8.19</checkstyle.version>
+    <checkstyle.version>8.20</checkstyle.version>
     <sonar.version>6.7</sonar.version>
     <sonar-java.version>5.12.0.17701</sonar-java.version>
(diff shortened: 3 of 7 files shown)


romani:
Please share some prove that new Check can do violation reporting ( any violation) to make sure basic testing was done.

muhlba91:
missingjavadoctype

romani:
please rebase on latest master, to make CI green.

muhlba91:
@romani - CI runs through except TC. TC still tries to look for `checkstyle-sonar-plugin/pom.xml` instead of `pom.xml` in the root directory - can you please adapt the configuration and re-run TC. :)

romani:
do we want to release 4.20 now , or should we wait something ?

muhlba91:
In my opinion nothing is against not releasing it now. I wouldn't see major/important issues open which should get fixed before a release. :) However, since `4.19` got released yesterday, the only reason for waiting, e.g., another week is in case issues pop up now as `4.19` provides support for the new `sonar-java` version.
What do you think?

romani:
lets wait for 1-2 week ...

please provide update to wiki page on new release process.

muhlba91:
perfect :)

I already updated the wiki page a few days ago. Steps 1-6 stay the same and there's only a step 7.1 and 7.2 now.

src/main/resources/com/sonar/sqale/checkstyle-model.xml
```diff
@@ -894,6 +894,19 @@
           <txt>min</txt>
         </prop>
       </chc>
+      <chc>
+        <rule-repo>checkstyle</rule-repo>
+        <rule-key>com.puppycrawl.tools.checkstyle.checks.javadoc.MissingJavadocTypeCheck</rule-key>
```
rnveach: @muhlba91 Does every module need to be listed here?
It doesn't seem to be part of our test to verify everything is listed in here. https://github.com/checkstyle/sonar-checkstyle/blob/master/src/test/java/org/sonar/plugins/checkstyle/internal/ChecksTest.java#L87-L94

muhlba91: I don't use SQALE on my instances but I don't think it's necessary since, afaik, if it's not listed iit doesn't have an impact on the score?
However, I added it because there were other Javadoc checks present and I kept it on-par with them. Do we have any rules for the SQALE stuff?

Anyhow, how is SQALE the SQALE support in SonarQube atm? At least in the API I read that it got deprecated:
- http://javadocs.sonarsource.org/7.7/apidocs/deprecated-list.html and search for SQALE
- https://jira.sonarsource.com/browse/SQALE-277?jql=project%20%3D%20SQALE -> the project is named "[DEPRECATED] SQALE"

rnveach: This is my first time seeing it so I don't believe there are any rules. I don't know what SQALE is so I couldn't identify what needs to be done with it.
Unless you or @romani know more, I suggest we just create an issue for a test to verify it has valid information and not require checks be listed.

romani: this config came from Sonarqube team.
I never understand how to it is possible to estimate fix in time, in most cases it does work at all. 
Such estimations NEVER worked well, it was only source of laugh ..... ones you spend a day with "10 min fix".
If Sonar do not demand such config be for each rule - I would be happy to remove them completely.

muhlba91: ah ok, I'll see if Sonar is happy when the file is empty/removed. I think the listed checks are anyways not up to date with all checks available.
If so, we can remove it and also don't have to copy over the SQALE source code from sslr. Will create an issue on that latest tomorrow. :)

muhlba91: created #219

Issue #220, opened 2019-05-14
RulePriority deprecated


muhlba91:
As per Sonar `RulePriority` is deprecated: https://github.com/SonarSource/sonarqube/blob/master/sonar-plugin-api/src/main/java/org/sonar/api/rules/RulePriority.java#L30

However, SonarQube itself still uses `RulePriority` as a return value, e.g. in their `ActiveRules`: https://github.com/SonarSource/sonarqube/blob/master/sonar-plugin-api/src/main/java/org/sonar/api/rules/ActiveRule.java#L120.

Hence, once SonarQube removes their usage we can remove the deprecated accesses to `RulePriority` as well.

Issue #219, opened 2019-05-13
How to handle SQALE?


muhlba91:
SonarQube seems to have deprecated SQALE:
- http://javadocs.sonarsource.org/7.7/apidocs/deprecated-list.html
- https://jira.sonarsource.com/browse/SQALE-277?jql=project%20%3D%20SQALE (see project name)

For now, we are using the `sqale.xml` file to define, although only partly, the SQALE ratings of each check. However, first, they are not defined well (and for some it might be difficult defining them) and, secondly, what would be the new SQ process for those ratings?

Furthermore, for reading this XML file we rely on the deprecated (and to be removed without replacement) class https://github.com/SonarSource/sslr-squid-bridge/blob/master/src/main/java/org/sonar/squidbridge/rules/SqaleXmlLoader.java#L42.
If SQALE is still supported and need, we must ensure to have a proper implementation working on our side to be independent of this library.

Hence, **check** if SQALE is still **mandatory** - if:
- **not**: remove `sqale.xml` and all related code
- **mandatory**: copy over implementation from library

See:
- https://github.com/checkstyle/sonar-checkstyle/pull/216#discussion_r279843386
- https://github.com/checkstyle/sonar-checkstyle/issues/217#issuecomment-490434067

muhlba91:
note: `sqale.xml` contains only an example after transitioning to loading via metadata.
to be investigated: do we need to set sqale still nowadays?

romani:
>do we need to set sqale still nowadays?

I never understand it. Let's drop it.

PR #221, opened 2019-05-14
[WIP] resolve SonarCloud violations

### LINKED ISSUE #217: Resolve violations from sonarcloud
https://sonarcloud.io/dashboard?id=checkstyle_sonar-checkstyle

lets try to resolve all violations in our code, if reasonable.

muhlba91:
fixes #217;
except violations relating to #219  and #220.

### DIFF
--- a/src/main/java/org/sonar/plugins/checkstyle/CheckstyleRulesDefinition.java
+++ b/src/main/java/org/sonar/plugins/checkstyle/CheckstyleRulesDefinition.java
@@ -21,14 +21,16 @@ package org.sonar.plugins.checkstyle;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.URL;
 import java.util.Objects;
 import java.util.Properties;
 
+import com.google.common.base.Charsets;
+import com.google.common.io.Resources;
 import org.sonar.api.ExtensionPoint;
 import org.sonar.api.batch.ScannerSide;
 import org.sonar.api.server.rule.RulesDefinition;
 import org.sonar.api.server.rule.RulesDefinitionXmlLoader;
-import org.sonar.squidbridge.rules.ExternalDescriptionLoader;
 import org.sonar.squidbridge.rules.SqaleXmlLoader;
 
 import com.google.common.annotations.VisibleForTesting;
@@ -61,12 +63,28 @@ public final class CheckstyleRulesDefinition implements RulesDefinition {
                 .getResourceAsStream(xmlRulesFilePath)) {
             ruleLoader.load(repository, resource, "UTF-8");
         }
-        ExternalDescriptionLoader.loadHtmlDescriptions(repository, htmlDescriptionFolder);
+        loadHtmlDescriptions(repository, htmlDescriptionFolder);
         try (InputStream resource = CheckstyleRulesDefinition.class
                 .getResourceAsStream("/org/sonar/l10n/checkstyle.properties")) {
             loadNames(repository, resource);
         }
         SqaleXmlLoader.load(repository, "/com/sonar/sqale/checkstyle-model.xml");
+    }
+
+    private static void loadHtmlDescriptions(NewRepository repository, String htmlDescriptionFolder) {
+        // code adapted from:
+        // https://github.com/SonarSource/sslr-squid-bridge/blob/2.7.0.377/
+        // src/main/java/org/sonar/squidbridge/rules/ExternalDescriptionLoader.java
+        for (NewRule rule : repository.rules()) {
+            final URL resource = CheckstyleRulesDefinition.class.getResource(htmlDescriptionFolder
+                    + "/" + rule.key() + ".html");
+            try {
+                rule.setHtmlDescription(Resources.toString(resource, Charsets.UTF_8));
+            }
+            catch(IOException e) {
+                throw new IllegalStateException("Failed to read: " + resource, e);
+            }
+        }
     }
 
     private static void loadNames(NewRepository repository, InputStream stream) {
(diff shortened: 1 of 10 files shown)


src/main/java/org/sonar/plugins/checkstyle/CheckstyleRulesDefinition.java
```diff
@@ -61,14 +63,30 @@ static void extractRulesData(NewRepository repository, String xmlRulesFilePath,
...
                 .getResourceAsStream("/org/sonar/l10n/checkstyle.properties")) {
             loadNames(repository, resource);
         }
         SqaleXmlLoader.load(repository, "/com/sonar/sqale/checkstyle-model.xml");
     }
 
+    private static void loadHtmlDescriptions(NewRepository repository, String htmlDescriptionFolder) {
+        // code adapted from:
+        // https://github.com/SonarSource/sslr-squid-bridge/blob/2.7.0.377/
+        // src/main/java/org/sonar/squidbridge/rules/ExternalDescriptionLoader.java
```
rnveach: Can we move these comments to the Javadoc?
Maybe copy the first sentence from sonar's javadoc (if there is one) and append a note saying it was copied from and use `{@link ExternalDescriptionLoader#loadHtmlDescriptions(NewRepository, String)}` instead of url to github.

muhlba91: Yes, but should we also move https://github.com/checkstyle/sonar-checkstyle/blob/master/src/main/java/org/sonar/plugins/checkstyle/CheckstyleRulesDefinition.java#L73 to the Javadoc as it follows the same.

However, I don't quite understand what you mean with `{@link ExternalDescriptionLoader#loadHtmlDescriptions(NewRepository, String)}`?
Because for now, this class still exists, just like `loadNames` (see link before) still existed until they removed some of their deprecated code. I assume this will happen here as well, meaning `ExternalDescriptionLoader` won't exist at some point anymore.

Both methods don't have any Javadoc in `sslr-squid-bridge`.

romani:
Please do not try to fix all problems in one commit and one PR. Please split the work to ease review

src/main/java/org/sonar/plugins/checkstyle/CheckstyleRulesDefinition.java
rnveach: > shouldn't we also move ... to the Javadoc as it follows the same?

Yes I like that.

> Because for now, this class still exists, just like loadNames (see link before) still existed until they removed some of their deprecated code. I assume this will happen here as well, meaning ExternalDescriptionLoader won't exist at some point anymore.

Ok, URL has tag in it so it will never go bad unless github goes away.
`{@link` was just a javadoc tag that would take user to that method if they clicked on it like in an IDE. If class/method does go away, javadoc tool should print an error on the `{@link` as it won't point to a valid class.

I was mainly just pointing its usage since I feel it is a better way to point to a class/method in a javadoc. If everyone wants to stay with URL, I am fine.

romani: @muhlba91 , it this discussion point resolved ?
If yes, please pin @rnveach to approve PR.

rnveach: It has not been resolved.

rnveach:
Code coverage is failing on travis.

src/main/java/org/sonar/plugins/checkstyle/CheckstyleRulesDefinition.java
muhlba91: For this point I was waiting for a decision if we want to put dead Javadoc links in or stick with GitHub URL references?
I'm good with both ways but we should still keep the URL somewhere as well for a full reference to where the code was taken from (incl. `sslr-squid-bridge` version). Otherwise, you won't really know from where/when the code was taken over and whether it was modified by us already or not.

muhlba91:
@romani sorry, these weeks, unfortunately, I'm quite busy. This is why development of this issue has been a bit stalled. I will think about how to split it into more commints and PR, so either per violation or per rule. I'm not entirely sure which one is more useful though? Then I'm going to close this PR but will go to #222 first.

romani:
>I'm not entirely sure which one is more useful though? 

it is up to you, it is my recommendation. Long/Big PRs are prone to stay in long review and usually one questionable code piece keep whole PR. So it is recommendation to do all step by step or in smaller pieces that easy to consume by reviewer and in case of not-simple cases other fixes are not blocked.

src/main/java/org/sonar/plugins/checkstyle/CheckstyleRulesDefinition.java
romani: I see no reason in reference of javadoc in other project, especially to non-existing code (non-existing javadoc).

Please move link from single comment to  loadHtmlDescriptions javadoc styles comment. Please do the same in "loadNames". Simply move whole text of comment to javadoc (no extra descriptions are required.)

romani:
@muhlba91 , ping.
It is better to resolve issues and merge sooner.

Please [move comment to javadoc](https://github.com/checkstyle/sonar-checkstyle/pull/221#discussion_r287826570) and lets merge this PR.

romani:
@muhlba91 , please do final update and lets merge this PR

romani:
@muhlba91 , can you finish this PR ? or should we close it ?

FYI: I updated checkstyle project to actually use Sonar as CI validator for PRs,  it would be awesome to eventually Sonar to control our code and in this plugin.

muhlba91:
i'll close it for now and work on it later.

Issue #189, opened 2019-01-27
upgrade to checkstyle 8.17


romani:
https://checkstyle.org/releasenotes.html#Release_8.17 , upgrade version at https://github.com/checkstyle/sonar-checkstyle/edit/master/pom.xml#L88

> RightCurly: remove obsolete shouldStartLine property

metadata need to be updated.

romani:
violations on 4.16 version - 24 violation, Rightcurly with `shouldStartLine=false`:
![image](https://user-images.githubusercontent.com/812984/51807569-245dfe80-223d-11e9-8682-be9fbb129123.png)

afgter upgrade:
```
[ERROR] Failed to execute goal org.sonarsource.scanner.maven:sonar-maven-plugin:3.4.1.1168:sonar (default-cli) on project checkstyle-sonar-plugin-parent: Can not execute Checkstyle: cannot initialize module TreeWalker - Property 'shouldStartLine' in module RightCurly does not exist, please check the documentation -> [Help 1]
```

romani:
fix is merged at #190,  commit is referencing wrong issue number.

romani:
RFF on sonarqube - https://community.sonarsource.com/t/rff-checkstyle-plugin-4-17/6192

romani:
https://community.sonarsource.com/t/new-release-checkstyle-plugin-4-17/6767

FelikZ:
@romani 
> create new RightCurly Rule, based on template of new plugin version, copy all configuration from old Rule.

Please, clarify what exactly you mean by `create new RightCurly Rule`? Where this available in UI?
Thanks!

Issue #124, opened 2017-08-31
upgrade to checkstyle 8.2


romani:
http://checkstyle.sourceforge.net/releasenotes.html#Release_8.2 , 
upgrade version at https://github.com/checkstyle/sonar-checkstyle/edit/master/pom.xml#L88

==============

FROM CHECKSTYLE RELEASE NOTES  that might affect plugin

Breaking backward compatibility:

    JavadocParser: inconsistent AST tree with and without SINGLETON_ELEMENT. 
    move AbstractTypeAwareCheck and ClassResolver to javadoc package. 
    remove deprecated property maxLineLength from LeftCurlyCheck. 
    remove FileContentsHolder module as FileContents object is available for filters on TreeWalker in TreeWalkerAudit Event.
    Replace terms in AST Tokens by terms from HTML specification.

New:

    new Check: single line annotation location.
    Add HTML5 tags support to Javadoc antlr4 grammar and related classes. 
    UnnecessaryParentheses: Reject parentheses around single parameters in lambdas. 
    Add token type to LocalizedMessage. 

**ATTENTION** (migration notes):
1)if happens:
`[ERROR] Failed to execute goal org.sonarsource.scanner.maven:sonar-maven-plugin:3.4.0.905:sonar (default-cli) on project checkstyle-sonar-plugin-parent: Can not execute Checkstyle: cannot initialize module TreeWalker - Property 'maxLineLength' in module LeftCurly does not exist, please check the documentation -> [Help 1]
`
solution is to remove "Custom Rule" and recreate it from "Rule Template"
old rule:
 ![image](https://user-images.githubusercontent.com/812984/34415597-7cd29e3e-eba4-11e7-972a-bed5abcd0491.png)
new rule should be without property:
![image](https://user-images.githubusercontent.com/812984/34415651-c20129ee-eba4-11e7-9e8b-41c4adc4fb37.png)

2) for UnnecessaryParentheses Rule, LAMBDA token has to added manually in UI or recreate rule from scratch.

tsjensen:
What is your current strategy with SonarQube? I see that the last release was for Checkstyle 7.7. For a time it seemed that you would be releasing a matching SonarQube plugin for every Checkstyle release. At which interval do you plan to make SonarQube plugin releases? Thanks for the info.

romani:
sorry , we just run out of time to support main project and this plugin, and to keep it in sync.
We had too much activity in main project.

[...]

PR #223, opened 2019-06-03
Issue #222: update to CS 8.21

### LINKED ISSUE #222: upgrade to checkstyle 8.21
https://checkstyle.org/releasenotes.html#Release_8.21 , upgrade version at https://github.com/checkstyle/sonar-checkstyle/edit/master/pom.xml#L88

Focus items:

>    NewlineAtEndOfFile default should be OS agnostic. 

we need to change default value of Check property

>    Remove PARAMETER_DEF from AnnotationLocation. 

chnages in token set

>    API: make api not depend on antlr by splitting DetailAST between implementation and interface.

no changes mostlikely is requied, just recheck that plugin works
 
>    new check: MissingJavadocMethod from JavadocMethod. 

new MissingJavadocMethod and removal of property from JavadocMethod

muhlba91:
- fixes #222 
- only check for presence of non-deprecated properties on checks


src/test/java/org/sonar/plugins/checkstyle/internal/ChecksTest.java
```diff
@@ -452,7 +452,7 @@ else if (AbstractFileSetCheck.class.isAssignableFrom(clss)) {
         final PropertyDescriptor[] map = PropertyUtils.getPropertyDescriptors(clss);
 
         for (PropertyDescriptor p : map) {
-            if (p.getWriteMethod() != null) {
+            if (p.getWriteMethod() != null && !p.getWriteMethod().isAnnotationPresent(Deprecated.class)) {
```
rnveach: @romani We don't hide deprecated properties in main repo. We should do that in both places?

romani: Can you explain me what is reason why we have should show deprecated users properties ?
Example: https://checkstyle.org/config_javadoc.html#JavadocMethod_Properties (ignoreMethodNamesRegex). If no good reason , we should remove them, to make breaking compatibility actual.

We should remove properties as clear signal that they are not effective any more and to not give user any fake filling that Check is working as before.
user should not get to know in hard way that property is not working any more, we should fail fast.

In main repo, we did not have Deprecated properties, if we deprecate we did removal of methods and removal from documentation.

rnveach: @romani 
>  If no good reason , we should remove them, to make breaking compatibility actual.

I thought this was the reason to mark them deprecated. To not break compatibility completely in 1 go, otherwise I can just remove the methods completely which would break every config using the module and require an override for our CI.

> In main repo, we did not have Deprecated properties, if we deprecate we did removal of methods and removal from documentation.

So remove all deprecated properties from main repo? If you agree, I will create an issue.

romani: If we break compatibility functionally it is the same as breaking it in config. If both is together, users will have to spend time with Checkstyle only ones. I almost 99% sure, nobody activate new Checks right after release. Majority of users do not care about our releases till something inconvenient happens on their side and and they search for fix by upgrade to latest version.

So, I do not think that migration process is required for properties of Check. It might be ok make it in few steps if we need it.

Please create new issue, it will be one more breaking change.

romani:
I am surprised  that I do not see my comment about testing artifacts .... 

@muhlba91 , please provide some testing artifacts/screenshots/logs to prove that you tested upgrade and new Check is executing without problems ..... and we can make a release.

muhlba91:
@romani sorry, your message got lost due to mail server failure in the company.

Please see screenshot of the check running through. :)
missingjavadocmethodcheck


FYI, I realized that I have set the default scope to private and not public -> fixed now.

src/main/resources/org/sonar/plugins/checkstyle/rules.xml
```diff
@@ -814,9 +814,6 @@
     <tag>comment</tag>
     <configKey><![CDATA[Checker/TreeWalker/JavadocMethod]]></configKey>
     <cardinality>MULTIPLE</cardinality>
-    <param key="minLineCount" type="INTEGER">
-      <defaultValue>-1</defaultValue>
-    </param>
```
romani: we need to update issue description with migration notes to let user know what to do when problem happen with Check after upgrade.

Please configure JavadocMethodCheck on already released version of plugin, run it, upgrade to new version of plugin, run again, experience problem, and we need to explain users how resolve it. I think user will have to recreate Rule from scratch.

please share exception in issue, to let users quickly find migration notes by such exception.

src/main/resources/org/sonar/plugins/checkstyle/rules.xml
```diff
@@ -1240,7 +1257,7 @@
     <name><![CDATA[Newline At End Of File]]></name>
     <configKey><![CDATA[Checker/NewlineAtEndOfFile]]></configKey>
     <param key="lineSeparator" type="s[system,crlf,cr,lf,lf_cr_crlf]">
-        <defaultValue>system</defaultValue>
+        <defaultValue>lf_cr_crlf</defaultValue>
```
romani: just mention in Mogration notes that without update of Rule, it will work with `system` value - rule need to be recreated or value is manually changed.

src/main/resources/org/sonar/plugins/checkstyle/rules.xml
```diff
@@ -1956,7 +1973,7 @@
     <param key="allowSamelineParameterizedAnnotation" type="BOOLEAN">
       <defaultValue>false</defaultValue>
     </param>
-    <param key="tokens" type="s[CLASS_DEF,INTERFACE_DEF,PACKAGE_DEF,ENUM_CONSTANT_DEF,ENUM_DEF,METHOD_DEF,CTOR_DEF,VARIABLE_DEF,PARAMETER_DEF,ANNOTATION_DEF,ANNOTATION_FIELD_DEF]">
+    <param key="tokens" type="s[CLASS_DEF,INTERFACE_DEF,PACKAGE_DEF,ENUM_CONSTANT_DEF,ENUM_DEF,METHOD_DEF,CTOR_DEF,VARIABLE_DEF,ANNOTATION_DEF,ANNOTATION_FIELD_DEF]">
```
romani: please prove that all existing instances of Rule that used such token will fail with exception, please share exception in issue, to let users quickly find migration notes by such exception.

romani:
@muhlba91 , ping.

src/main/resources/org/sonar/plugins/checkstyle/rules.xml
muhlba91: I thought already this won't fail. The user can't configure checks with the old/deleted properties but current checks won't fail because the underlying CS still accpets those properties.
Interestingly, which I wasn't aware of, SQ seems to save a copy of the old template for every check, meaning I can actually reconfigure my currently existing check with all the old properties but just can't create a new one using those removed properties.
However, I also couldn't get this old check raising any violations in my example code although with the previous SQ plugin version I have got some.
How should we declare and resolve this "issue"?

romani: > because the underlying CS still accpets those properties.

oh , yes, such removal was non standard, as original Check was not changed. Yes, it will continue to work till we actually remove properties from Check class.

> However, I also couldn't get this old check raising any violations in my example code although with the previous SQ plugin version I have got some.

Do you mean:
You create configure JavadocMethod on 8.20, and it works, show some violation.
After upgrade to 8.21, it stops reporting violations on the same places ?

Its is probably because we removed from Check all functionality that was moved to another Check - https://github.com/checkstyle/checkstyle/commit/f094c67d6c042a0f786665034b5dddbb665b955c#diff-01491bb8516b661ce590f3bc79015eedR173
looks like `setAllowMissingParamTags` still in this Check, try to play with it in confg , it should still be in action.

src/main/resources/org/sonar/plugins/checkstyle/rules.xml
muhlba91: issue created: https://github.com/checkstyle/sonar-checkstyle/issues/228

src/main/resources/org/sonar/plugins/checkstyle/rules.xml
muhlba91: where should we add those migration notes?

romani: I usually put them at bottom of upgrade issue description or as comment.
examples:
https://github.com/checkstyle/sonar-checkstyle/issues/189
https://github.com/checkstyle/sonar-checkstyle/issues/124

src/main/resources/org/sonar/plugins/checkstyle/rules.xml
romani: I did not mean new issue creation, just sharing exception with workaround in #222 should be enough

src/main/resources/org/sonar/plugins/checkstyle/rules.xml
muhlba91: Yes, exactly this is what I have observed. With 8.20 it reported violations and with 8.21 none were reported and your explanation makes totally sense. If the functionality is removed but all properties still "there" (but deprecated) it won't fail but also not report anything anymore.

src/main/resources/org/sonar/plugins/checkstyle/rules.xml
muhlba91: added comment: https://github.com/checkstyle/sonar-checkstyle/issues/222#issuecomment-502663551

src/main/resources/org/sonar/plugins/checkstyle/rules.xml
muhlba91: ah ok. well, should I remove the issue and copy it into #222 or do we link #222 with #228 and close #228?

src/main/resources/org/sonar/plugins/checkstyle/rules.xml
romani: Just put in mutation notes, most users are not aware that sonar plugin and checkstyle are different projects

src/main/resources/org/sonar/plugins/checkstyle/rules.xml
romani: Keep it, not a problem. Just do not create similar next time

src/main/resources/org/sonar/plugins/checkstyle/rules.xml
muhlba91: Perfect, I just created a migration comment: https://github.com/checkstyle/sonar-checkstyle/issues/222#issuecomment-503264472

Issue #225, opened 2019-06-15
resolve IDEA 2019.1.3 violations


romani:
build failure: https://teamcity.jetbrains.com/viewLog.html?buildId=2364830&tab=Inspection&buildTypeId=Checkstyle_SonarCheckstyleIdeaInspectionsMaster

HTML
  Deprecated HTML attribute (20)
src/main/resources/org/sonar/l10n/checkstyle/rules/checkstyle
com.puppycrawl.tools.checkstyle.checks.annotation.PackageAnnotationCheck.html (2)
com.puppycrawl.tools.checkstyle.checks.AvoidEscapedUnicodeCharactersCheck.html (5)
com.puppycrawl.tools.checkstyle.checks.design.DesignForExtensionCheck.html (4)
com.puppycrawl.tools.checkstyle.checks.indentation.CommentsIndentationCheck.html (4)
com.puppycrawl.tools.checkstyle.checks.naming.AbbreviationAsWordInNameCheck.html (5)

HTML
  Presentational HTML tag (Errors) (6)
src/main/resources/org/sonar/l10n/checkstyle/rules/checkstyle
com.puppycrawl.tools.checkstyle.checks.regexp.RegexpCheck.html (6)
22:  Presentational HTML tag 
22:  Presentational HTML tag 
42:  Presentational HTML tag 
42:  Presentational HTML tag 
51:  Presentational HTML tag 
51:  Presentational HTML tag 

JUnit issues
  Simplifiable JUnit assertion (Errors) (1)
src/test/java/org/sonar/plugins/checkstyle/internal
ChecksTest.java (1)
90: verifyTestConfigurationFiles() assertTrue() can be simplified to 'assertFalse()' 

 ==============

https://stackoverflow.com/a/16974491/1015848
https://softwareengineering.stackexchange.com/a/255588

romani:
fix is merged.

PR #226, opened 2019-06-15
Issue #225: resolve Idea vilolations on deprecated HTML tags

romani:
#225

### DIFF
--- a/src/main/resources/org/sonar/l10n/checkstyle/rules/checkstyle/com.puppycrawl.tools.checkstyle.checks.AvoidEscapedUnicodeCharactersCheck.html
+++ b/src/main/resources/org/sonar/l10n/checkstyle/rules/checkstyle/com.puppycrawl.tools.checkstyle.checks.AvoidEscapedUnicodeCharactersCheck.html
@@ -1,8 +1,8 @@
 <div >
-<h2><a name="AvoidEscapedUnicodeCharacters"></a>AvoidEscapedUnicodeCharacters</h2>
+<h2><a id="AvoidEscapedUnicodeCharacters"></a>AvoidEscapedUnicodeCharacters</h2>
       
 <div >
-<h3><a name="Description"></a>Description</h3>
+<h3><a id="Description"></a>Description</h3>
         
 <p>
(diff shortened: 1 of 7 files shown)


src/main/resources/org/sonar/l10n/checkstyle/rules/checkstyle/com.puppycrawl.tools.checkstyle.checks.AvoidEscapedUnicodeCharactersCheck.html
```diff
@@ -1,8 +1,8 @@
 <div >
-<h2><a name="AvoidEscapedUnicodeCharacters"></a>AvoidEscapedUnicodeCharacters</h2>
+<h2><a id="AvoidEscapedUnicodeCharacters"></a>AvoidEscapedUnicodeCharacters</h2>
```
rnveach: @romani I feel this change broke the original reason for this tag.
This code was similar to what the generated source of our xdocs look like.
`name` attribute of `a` allows you to place links to these sections similar to http://checkstyle.sourceforge.net/config_filters.html#SuppressionXpathSingleFilter .
Switching it to `id` serves absolutely no purpose if you aren't referencing this tag with javascript or css because `id` is just a unique identifier to identify this tag out of all of the ones on the page.

IMO, code like this should be reverted, or removed completely.

romani: You can use Id after `#` , links will work.
Content of such pages are so outdated, so investing time in them not is valuable. I just tried to make CI green.
Doc in plugin should have only short description and link to our site.

Looks like we have the same violations in main project, so

rnveach: Looks like you are right, it changed for new HTML versions (HTML 5).
https://stackoverflow.com/a/484781/1016482
Old browsers won't support `id`.

PR #255, opened 2019-12-09
Issue #253: update to CS 8.27

### LINKED ISSUE #253: upgrade to checkstyle 8.27
https://checkstyle.org/releasenotes.html#Release_8.27 , upgrade version at https://github.com/checkstyle/sonar-checkstyle/edit/master/pom.xml#L88

Plugin is affected by new check introduction only

muhlba91:
fix #253 

I have got a problem here as the new check doesn't seem to work.

My example Java class is:
```
/** This comment causes a violation because it starts from the first line
  * and spans several lines.
  */
/**
  * This comment is OK because it starts from the second line.
  */
/** This comment is OK because it is on the single line. */
public class App {
    /** This comment causes a violation because it starts from the first line
      * and spans several lines.
      */
    /**
      * This comment is OK because it starts from the second line.
      */
    /** This comment is OK because it is on the single line. */
    public String getGreeting() {
        return "Hello world.";
    }

    /** This comment causes a violation because it starts from the first line
      * and spans several lines.
      */
    /**
      * This comment is OK because it starts from the second line.
      */
    /** This comment is OK because it is on the single line. */
    public static void main(String[] args) {
        System.out.println(new App().getGreeting());
    }
}
```
which is the example shown here: https://checkstyle.sourceforge.io/config_javadoc.html#JavadocContentLocation_Examples

As for the check in SQ, this is my settings (default values according to https://checkstyle.sourceforge.io/config_javadoc.html#JavadocContentLocation_Properties:
javadoccontentlocation

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ Compatibility matrix from checkstyle team:
 
 Checkstyle Plugin|Sonar min|Sonar max|Checkstyle|Jdk
 -----------------|---------|---------|----------|---
+4.27|6.7  |7.7+|8.27|1.8
 4.26|6.7  |7.7+|8.26|1.8
 4.25|6.7  |7.7+|8.25|1.8
 4.24|6.7  |7.7+|8.24|1.8
--- a/pom.xml
+++ b/pom.xml
@@ -90,7 +90,7 @@
   </ciManagement>
 
   <properties>
-    <checkstyle.version>8.26</checkstyle.version>
+    <checkstyle.version>8.27</checkstyle.version>
     <sonar.version>6.7</sonar.version>
     <sonar-java.version>5.12.0.17701</sonar-java.version>
     <maven.checkstyle.plugin.version>2.17</maven.checkstyle.plugin.version>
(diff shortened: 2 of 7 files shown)


romani:
works fine on my side, just followed https://github.com/checkstyle/sonar-checkstyle/wiki/How-to-build-test-plugin-on-localm :
```
~/java/github/checkstyle/sonar-checkstyle [muhlba91/issue/253 L|✚ 1] 
$ git diff
diff --git a/src/main/java/org/sonar/plugins/checkstyle/rule/ActiveRuleWrapper.java b/src/main/java/org/sonar/plugins/checkstyle/rule/ActiveRuleWrapper.java
index 4ce5566..f30c7dd 100644
--- a/src/main/java/org/sonar/plugins/checkstyle/rule/ActiveRuleWrapper.java
+++ b/src/main/java/org/sonar/plugins/checkstyle/rule/ActiveRuleWrapper.java
@@ -21,8 +21,7 @@ package org.sonar.plugins.checkstyle.rule;
 
 import java.util.Map;
 
-/**
- * Due to our usage of {@link org.sonar.plugins.checkstyle.CheckstyleProfileExporter}
+/** Due to our usage of {@link org.sonar.plugins.checkstyle.CheckstyleProfileExporter}
  * on the {@link org.sonar.api.batch.ScannerSide} for generating CheckStyle configurations,
  * we must make it compatible for the new {@link org.sonar.api.batch.rule.ActiveRules}.
  *

```

image

PR #257, opened 2019-12-23
minor: formatting and fix url to checkstyle.org

romani:
wrap long line to be less than 100 symbols


romani:
we validate by maven checkstyle plugin - https://github.com/checkstyle/sonar-checkstyle/blob/master/pom.xml#L304
so it is files under `src` only - https://github.com/checkstyle/sonar-checkstyle/blob/master/pom.xml#L331

PR #267, opened 2020-01-20
replace cobertura with jacoco

muhlba91:
replacing cobertura with jacoco.
i also checed out the main project and found the way how to change the coverage on a per-file basis. hence, it's a conversion from cobertura to jacoco keeping the main configuration.


muhlba91:
merge before #268

romani:
Already merged as part of other pr

PR #268, opened 2020-01-20
JDK 11

muhlba91:
update to JDK 11;

merge after #267 !

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,11 +22,11 @@ branches:
 matrix:
   fast_finish: true
   include:
-    - jdk: openjdk8
+    - jdk: openjdk11
       env: CMD="./.ci/travis.sh install"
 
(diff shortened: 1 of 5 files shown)


romani:
We need to teach sonar on how to see jacoco coverage

muhlba91:
@romani wercker does the sonarcloud.io integration, right?
afaik, sonar needs to know where to find the `jacoco.exec` but this is only created by running the `check`s in maven incl. tests which are disabled in wercker as i saw. we need to run the tests in wercker as well, then `target/jacoco.exec` is created which should either be picked up by sonar anyways by default or a property has to be set.
can you enable the tests and instrumentation in wercker? (tests plus jacoco instrumentation are needed, we can skip checkstyle, pmd, spotbugs, ...)

Issue #269, opened 2020-01-20
Apply checkstyle verification for 'it' sources


romani:
from https://github.com/checkstyle/sonar-checkstyle/pull/258#discussion_r365800709

for some reason out `it` sources are not validated by checkstyle config. This need to be fixed.

muhlba91:
can you verify that the normal test sources are validated?
i cannot get a failure if i change any test file to either violate the line length or import order which fails immediately for any source file in 'src/main/java' .

PR #272, opened 2020-01-21
#269: Enable CS for test sources

muhlba91:
enables checkstyle for test sources as it was disabled by default.

i face one problem in `ChecksTest::getFinalProperties` (L400) with the lambda expression again:
- checkstyle demands braces, IDEA inspections don't
- not adding a line break here will result in a line being > 100 characters -> checkstyle fails

@romani please take a look


romani:
Please rebase and do not update pom.xml for now to let build pass and let me merge all other updates.

Please send separate PR for pom.xml update, I will take take a look.

muhlba91:
@romani done, see #273

PR #279, opened 2020-01-22
Issue #278: disable inspection that suggest to convert block to expression

### LINKED ISSUE #278: Resolve violation from inspection about lambda expression
https://teamcity.jetbrains.com/viewLog.html?buildId=2729988&tab=Inspection&buildTypeId=Checkstyle_SonarCheckstyleIdeaInspectionsMaster

```
src/test/java/org/sonar/plugins/checkstyle/internal
ChecksTest.java (1)
401: getFinalProperties() Statement lambda can be replaced with expression lambda
```

rule is not configurable, 
image

and it is disabled in Checkstyle repo already to not conflict with checkstyle rule.

romani:
fixes #278

### DIFF
--- a/config/intellij-idea-inspections.xml
+++ b/config/intellij-idea-inspections.xml
@@ -349,7 +349,7 @@
   <inspection_tool class="CloneableImplementsClone" enabled="false" level="ERROR" enabled_by_default="false">
     <option name="m_ignoreCloneableDueToInheritance" value="false" />
   </inspection_tool>
-  <inspection_tool class="CodeBlock2Expr" enabled="true" level="ERROR" enabled_by_default="true" />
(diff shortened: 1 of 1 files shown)


rnveach:
@romani Should we just copy config from main repo and remove new violations and set them to be fixed in a future issue/pr?

romani:
it is better to share the same files , but it is not possible to do as there are issues in sonar plugin that we never had time to resolve. In core project we more diligent about code quality so code project is using a way more strict configuration.
All this due to lack of human resources to resolve all violations.
I hope @muhlba91 can help this project to become in better form... but I still you recommend to resolve all sonar violations first, as more visible set of technical dept.

Issue #282, opened 2020-01-29
SonarQube / Java compatibility


clarkie82:
Sonar version: 7.9.1
Checkstyle sonar plugin version: 4.28
Java Sonar server version: 1.11.0_5
Java Sonar scanner version: 1.8.0_232

Steps to reproduce or description of problem:
Run "maven sonar:sonar"
This should work, but throws errors as the only supported JDK version is 11. Should this only be "compatible" with higher level SonarQube versions that are enforcing Java 11 for scanners

romani:
>Checkstyle sonar plugin version: 4.28

It is compiled on jdk 11

>Should this only be "compatible" with higher level SonarQube versions that are enforcing Java 11 for scanners

Looks like we missed something, we thought that 7.9 is only running on  jdk 11. Can you share details of what version sonar can be executed on jdk 8 ?

clarkie82:
The server is running version 11, but the scanners can be version 8 still which is when this plugin causes issues as the scanner throws errors. Downgrading to 4.27 fixes this

BerndFarkaDyna:
the same problem on my side (but running with gradle)

the build is failing with: 

> The plugin [checkstyle] does not support Java 1.8.0_222

but the sonar-runner says a few lines before

> SonarScanner will require Java 11 to run starting in SonarQube 8.x

romani:
@muhlba91 ,

from https://github.com/checkstyle/sonar-checkstyle/pull/258#issuecomment-575579814

> i got the file running and executing but i had to use jdk 11 since SQ 7.9 requires it.

did we miss something with where jdk11 is required ?
If scanner download plugin jar from server ... ups ... we upgraded to jdk11 a bit early. Please confirm.

muhlba91:
i had some time today to investigate this. SQ 7.9 server needs JDK 11 and, since we adapted integration (and IT) to 7.9 server, we had to run the verification pipeline using JDK 11.

what i missed was the requirements regarding sonar scanner which are - and i don‘t know why - somehow different. i never realized it because i‘m running all „everyday“ projects with JDK 11.
sonar scanner (gradle at least) apparently supports JDK 8 as long as the SQ server‘s version is not 8.x. because when running gradle it shows a warning that from SQ 8.x onwards the scanner requires JDK 11 as @BerndFarkaDyna posted.
anyhow, with SQ 7.9 it still works using JDK 8. this is, in my opinion, quite confusing to have a divergence here.
[...]

PR #283, opened 2020-01-29
#282 set compatibility to 1.8

BerndFarkaDyna:
#282

## just a test to try if it is possible to set target of the bytecode to 1.8 and keep everything on java 11

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,10 @@ matrix:
   include:
     - jdk: openjdk11
       env: CMD="./.ci/travis.sh install"
-
+   
+    - jdk: openjdk8
+      env: CMD="./.ci/travis.sh install"
+    
     - jdk: openjdk11
       env: CMD="./.ci/travis.sh integration-tests"
 
--- a/pom.xml
+++ b/pom.xml
@@ -103,7 +103,7 @@
       8.28
     </maven.sevntu.checkstyle.plugin.checkstyle.version>
     <maven.jacoco.plugin.version>0.8.5</maven.jacoco.plugin.version>
-    <java.version>11</java.version>
+    <java.version>1.8</java.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <!-- we disable ITs by default as it need download of 200MB sonar application jar -->
     <skipITs>true</skipITs>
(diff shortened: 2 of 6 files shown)


muhlba91:
looks good so far. :+1:

please also try changing the CI to use jdk 8 (except for the travis integration test one).

BerndFarkaDyna:
> looks good so far. 👍
> please also try changing the CI to use jdk 8 (except for the travis integration test one).

sure, but do not test the corner case we are failing then..

since if i change the CI to jdk8 we build with 8 and test with 8 which works…
but I'll give it a try

muhlba91:
sure :)
my reasoning is rather to see whether we can run unit tests with jdk 8 and only use jdk 11 for integration testing or if we then fail due to jdk specifics.

either way, from my point of view, as long as @romani doesn‘t see a blocker, we can build and test solely with jdk 11 as well but keep compatibility at jdk 8.

romani:
> we can build and test solely with jdk 11 as well but keep compatibility at jdk 8.

I am ok, we will unlikely to use jdk11 specfic features in recent future.

BerndFarkaDyna:
unfortunally the test code is seems to use an API which is only available on JDK11..

but anyway; i try to give it a fix

muhlba91:
> I am ok, we will unlikely to use jdk11 specfic features in recent future.

> but anyway; i try to give it a fix

awesome, thanks a lot! :)

btw, as long as i didn‘t mix anything up, this commit contains all fixes for jdk 11: https://github.com/checkstyle/sonar-checkstyle/commit/3195b481378157ed59e7eda912e69e08035a41ed

BerndFarkaDyna:
@muhlba91 , @romani : I made the unit tests compile/work against java8; and squashed my changes, please have a look on my PR and provide some feedback, it would be really great if this could get into master and be released soon..

THX 
Bernd

romani:
@BerndFarkaDyna ,

please update commit message to be `Issue #282: ...... ` it is good to have commit clearly reference to issue. Issue reference PR, ... so even in few years we will easily review whole context of WHY update was done in such way.

@muhlba91 , please reheck on your local, it is critical that you continue to do executions/releases on your local.

BerndFarkaDyna:
@romani  / @muhlba91 : changed the comment.. so everything should be ready for a merge

romani:
@muhlba91 , in core project there is UT that do validation of commit message to follow required format.
you can copy it plugin, this will let you do release without manual modification of release notes.

PR #287, opened 2020-03-09
Issue #286: update CS 8.30

### LINKED ISSUE #286: upgrade to checkstyle 8.30
https://checkstyle.org/releasenotes.html#Release_8.30

muhlba91:
fix #286

Furthermore, there are some errors in the documentation at https://checkstyle.sourceforge.io/config_coding.html#AvoidDoubleBraceInitialization.
Each `new ArrayList<Object>` i smissing braces: `new ArrayList<Object>()` will be correct.

Sonarqube:
doublebraces

### DIFF
--- a/src/it/java/org/checkstyle/plugins/sonar/CheckJarTest.java
+++ b/src/it/java/org/checkstyle/plugins/sonar/CheckJarTest.java
@@ -26,7 +26,7 @@ import java.io.File;
 import org.junit.Test;
 
 public class CheckJarTest {
-    private static final String VERSION = "4.29";
+    private static final String VERSION = "4.30";
 
     @Test
     public void testJarPresence() {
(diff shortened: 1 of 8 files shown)


romani:
> Furthermore, there are some errors in the documentation

do not hesitate to open issue on us to fix problems.
done: https://github.com/checkstyle/checkstyle/issues/7811

src/it/java/org/checkstyle/plugins/sonar/CheckJarTest.java
```diff
@@ -26,7 +26,7 @@
 import org.junit.Test;
 
 public class CheckJarTest {
-    private static final String VERSION = "4.29";
+    private static final String VERSION = "4.30";
```
romani: very unfortunate copy-paste ..... do we know how to avoid this version definition here ?
can we parse pom to get this value ? or .... ?

muhlba91: I created #288. I actually believe for this certain use case in the test, we could simply perform a file matching in the folder for the pattern. In CI there should really never be more than one jar file matching the name.

romani: Good idea

PR #289, opened 2020-03-18
Issue #288: match jar file name for CheckJarTest

### LINKED ISSUE #288: CheckJarTest should not rely on manual version update
In https://github.com/muhlba91/sonar-checkstyle/blob/master/src/it/java/org/checkstyle/plugins/sonar/CheckJarTest.java#L29 we rely on checking whether the jar file with a certain version got created.

This needs to be manually edited every time. Change to either automated detection of the version vi a `pom.xml` or changing the method to retrieve any jar file matching the name. Usually, especially on CI, only one jar file should be existing and not more (as opposed to some manual local testing).

muhlba91:
fix #288

### DIFF
--- a/src/it/java/org/checkstyle/plugins/sonar/CheckJarTest.java
+++ b/src/it/java/org/checkstyle/plugins/sonar/CheckJarTest.java
@@ -19,22 +19,30 @@
 
 package org.checkstyle.plugins.sonar;
 
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
 
-import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.List;
+import java.util.function.BiPredicate;
+import java.util.stream.Collectors;
 
 import org.junit.Test;
 
 public class CheckJarTest {
-    private static final String VERSION = "4.30";
+    private static final String MATCHER = ".*checkstyle-sonar-plugin-\\d+\\.\\d+(-SNAPSHOT)?\\.jar";
 
     @Test
-    public void testJarPresence() {
-        final boolean snapshotExists = new File("target/checkstyle-sonar-plugin-"
-                                                + VERSION + "-SNAPSHOT.jar").exists();
-        final boolean releaseExists = new File("target/checkstyle-sonar-plugin-"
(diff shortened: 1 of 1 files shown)


src/it/java/org/checkstyle/plugins/sonar/CheckJarTest.java
```diff
@@ -19,22 +19,27 @@
...
-        final boolean snapshotExists = new File("target/checkstyle-sonar-plugin-"
-                                                + VERSION + "-SNAPSHOT.jar").exists();
-        final boolean releaseExists = new File("target/checkstyle-sonar-plugin-"
-                                               + VERSION + ".jar").exists();
-        assertTrue("Jar should exists",
-                   snapshotExists || releaseExists);
+    public void testJarPresence() throws IOException {
+        final List<Path> files = Files.find(Paths.get("target"), 1, (path, basicFileAttributes) -> {
+            return path.toString()
+                    .matches(".*checkstyle-sonar-plugin-\\d+\\.\\d+(-SNAPSHOT)*\\.jar");
```
rnveach: > (-SNAPSHOT)*

Shouldn't this be `(\\-SNAPSHOT)?` ?

muhlba91: `?` is a good catch!
As for escaping the `-` it should not need escaping due to the grouping. The test runs through with normal and snapshot jars, no matter if I escape or not.

src/it/java/org/checkstyle/plugins/sonar/CheckJarTest.java
```diff
@@ -19,22 +19,27 @@
...
-        final boolean releaseExists = new File("target/checkstyle-sonar-plugin-"
-                                               + VERSION + ".jar").exists();
-        assertTrue("Jar should exists",
-                   snapshotExists || releaseExists);
+    public void testJarPresence() throws IOException {
+        final List<Path> files = Files.find(Paths.get("target"), 1, (path, basicFileAttributes) -> {
+            return path.toString()
+                    .matches(".*checkstyle-sonar-plugin-\\d+\\.\\d+(-SNAPSHOT)?\\.jar");
+        })
+                .collect(Collectors.toList());
```
romani: please move lambda out to have a variable for it.
it is very hard to read code like this.

rnveach:
@muhlba91 CI is failing. Checkstyle errors.

Issue #87, opened 2017-03-17
Add UT validation of cardinality in rules


rnveach:
https://github.com/checkstyle/sonar-checkstyle/blob/1862ce2e39c548947ac902c01ff3551bf0bc4722/checkstyle-sonar-plugin/src/main/resources/org/sonar/plugins/checkstyle/rules.xml#L135

Looking at checks that have MULTIPLE carnality and don't, I believe it allows users to create multiple instances of the check giving them the `create rule` dialog. It requests all the options of the check set plus a new name and description.

To make sure it is set correctly (on or off), we should validate the cardinality to make sure it is correct.
We should set it to `MULTIPLE` if the check has more than 2 properties (including `tokens`).

Reasoning:
No options is self evident.
If just has token, user can specify how ever he wants. Multiple instances can just be combined into 1 with no difference.
If it has 2 options or more, user could want different values for each. Example: `RightCurlyCheck` could want `alone` for tokens `A,B,C` but want `same` for tokens `X,Y,Z`.

rnveach:
https://github.com/SonarSource/sonarqube/blob/8759b3933d8bc4a147b7242e1838a39b82bcc719/sonar-plugin-api/src/main/java/org/sonar/api/server/rule/RulesDefinitionXmlLoader.java#L100-L101
Shows `MULTIPLE` cardinality equals it is a "Template" rule.

romani:
example of Check that should not have multiple instances and has no properties that user can customize - http://checkstyle.sourceforge.net/config_coding.html#NoClone
 (there are more in codding group of Checks)

muhlba91:
closing this as it should not be necessary anymore after moving to metadata mechanism. we do not use translations anymore in this project.

PR #297, opened 2020-06-15
Issue #296: PoC Changes for Checkstyle Metadata Integration - meta up…

### LINKED ISSUE #296: PoC Changes for Checkstyle Metadata Integration
### Changes to introduce

- [x] Create utility to override Repository with metadata values
- [x] Create utility to create new Rules in the Repository if they are not present in `rules.xml`(it becomes kind of _freezed_ when these are merged in master)
- [x] For the above step, write a method to store plugin specific values in YAML.
- [x] Add dynamic loading for acceptable tokens
- [x] Add dynamic loading for enum values

gaurabdg:
Resolves #296 

- [x] Create utility to override Repository with metadata values
- [x] Create utility to create new Rules in the Repository if they are not present in `rules.xml`(it becomes kind of _freezed_ when these are merged in master)
- [x] For the above step, write a method to store plugin specific values in YAML.
- [x] Add dynamic loading for acceptable tokens
- [x] Add dynamic loading for enum values 

### PoC that the checkstyle meta produces same results:

`EmptyCatchBlock` present in `rules.xml`
Screenshot_20200619_022439

`EmptyCatchBlock` removed from `rules.xml`
Screenshot_20200619_022603

### Rule Description

`EmptyCatchBlock` description before update
before

`EmptyCatchBlock` description after update
after

### Tag Integration

The new tag assgining feature is only for the checks not present in rules.xml, i.e. new Checks. I removed three rules to generate these screenshots.
Screenshot_20200704_121701
Screenshot_20200704_121723
Screenshot_20200704_123043


romani:
@gaurabdg , as second commit (to ease revert of it) please remove all data about this check from plugin, to make sure we can substitude all required data by plugin. Please execute plugin and share details on how it works.
Please read wiki pages - https://github.com/checkstyle/sonar-checkstyle/wiki they describe how to test sonar+plugin on local by docker. Please let us know if you run into any problems.

pom.xml
```diff
@@ -288,8 +288,35 @@
...
+      <groupId>org.example</groupId>
+      <artifactId>metadata-gen</artifactId>
+      <version>1.0</version>
+    </dependency>
   </dependencies>
 
+  <repositories>
+    <repository>
+        <id>local-maven-repo</id>
+        <url>file://${project.basedir}/lib</url>
```
romani, line 316: Please install metadata-gen to local repo by `mvn install` so you can use it without this section.
You can update CI to clone latest from metadata-gen repo, install in to local repo of CI and build plugin only after.

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckUtil.java
```diff
@@ -0,0 +1,207 @@
...
+     * @throws IOException if the attempt to read class path resources failed.
+     * @see #isCheckstyleModule(Class)
+     */
+    public static Set<Class<?>> getCheckstyleModules() throws IOException {
+        final Set<Class<?>> checkstyleModules = new HashSet<>();
+
+        final ClassLoader loader = Thread.currentThread()
+                .getContextClassLoader();
+        final ClassPath classpath = ClassPath.from(loader);
+        final String packageName = "com.puppycrawl.tools.checkstyle";
```
romani, line 54: What about sevntu ?

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadata.java
```diff
@@ -0,0 +1,280 @@
...
+            throw new IllegalStateException("exception occured during loadMeta", ex);
+        }
+        return moduleDetails;
+    }
+
+    public void updateRulesWithMetadata() {
+        repository.rules().forEach(rule -> {
+            final String checkName = rule.key().substring(rule.key().lastIndexOf('.') + 1);
+            // condition required only for PoC
+            if ("EmptyBlockCheck".equals(checkName) || "JavadocPackageCheck".equals(checkName)) {
```
romani, line 79: There should be no such checks, we should load all that is acceptable by class path.

src/main/resources/org/sonar/plugins/checkstyle/metadata/additional-rules-meta.yml
```diff
@@ -0,0 +1,10 @@
+rules:
+    - rule: EmptyBlockCheck
+      priority: MAJOR
+#      tag:
+      status: READY
+#      cardinality: # should we keep it here? It's not loaded into repository rule.
```
romani, line 6: We need to auto figure out this, if we have configurable properties, so we can have few instances of Check.

pom.xml
```diff
@@ -580,7 +608,7 @@
                     <limit>
                       <counter>LINE</counter>
                       <value>COVEREDRATIO</value>
-                      <minimum>0.89</minimum>
+                      <minimum>0.87</minimum>
```
muhlba91, line 611: what is the reason to reduce it?
@romani wasn't the problem that SQ needs a passed quality gate and they also check the coverage?

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadata.java
```diff
@@ -0,0 +1,280 @@
...
+                    valuesArray, valuesArray.length, String[].class)));
+        }
+        else {
+            param.setType(MAP_PARAM_VALUES.get(paramType));
+        }
+    }
+
+    private static String convertName(String name) {
+        final int capacity = 1024;
+        final StringBuilder result = new StringBuilder(capacity);
```
muhlba91, line 173: StringBuilder or StringBuffer does not need a capacity.
please use StringBuffer as it's also thread-safe.

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadata.java
```diff
@@ -0,0 +1,280 @@
...
+            final Object[] valuesArray = getEnumValues(paramType).toArray();
+            param.setType(RuleParamType.singleListOfValues(Arrays.copyOf(
+                    valuesArray, valuesArray.length, String[].class)));
+        }
+        else {
+            param.setType(MAP_PARAM_VALUES.get(paramType));
+        }
+    }
+
+    private static String convertName(String name) {
```
muhlba91, line 171: please add a comment to what this one does with an example.
it's not really clear from the code to me...

src/main/resources/org/sonar/plugins/checkstyle/metadata/additional-rules-meta.yml
muhlba91: please detect it automatically. it's not additional-rules to be specified by the maintainer but should be detected whenloading and converting metadata

src/main/resources/org/sonar/plugins/checkstyle/rule-deletedChecks.xml
```diff
@@ -0,0 +1,2255 @@
+<rules>
```
muhlba91, line 1: why do we need this file?

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadata.java
muhlba91, line 79: are those new checks or checks to be updated? are those "complex" checks covering all scenarios of checks?
i'd like to have a PoC covering as many use cases/permutations as possible. :)

pom.xml
gaurabdg: I've used `mvn-install` plugin now. Do I also  need to also clone latest?

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckUtil.java
gaurabdg: I thought of preparing a PoC model first  with checkstyle checks and get mentors' approval and then tweak further.

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadata.java
gaurabdg: Done.

src/main/resources/org/sonar/plugins/checkstyle/metadata/additional-rules-meta.yml
gaurabdg: How will we get this from checks/meta? For example `tag` is plugin specific and always set by the maintainer I guess, during each check addition?

pom.xml
gaurabdg: My bad, I have rectified it.

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadata.java
gaurabdg: Well, TeamCity inspections are giving code improvement suggestions that `StringBuffer needs capacity` and `StringBuffer should be replaced by StringBuilder`. Shall I ignore them?

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadata.java
gaurabdg: Done

src/main/resources/org/sonar/plugins/checkstyle/rule-deletedChecks.xml
gaurabdg: It is for testing, whether `createMeta` matches default rule creator.

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadata.java
gaurabdg: I have supplid quite a number of checks to form almost all possible use cases I can think of.

src/main/resources/org/sonar/plugins/checkstyle/rule-deletedChecks.xml
romani: Please provide more details on this

romani: Please provide more details on this

romani: Please provide more details on this

.ci/travis.sh
```diff
@@ -5,6 +5,7 @@ set -euo pipefail
 case "$1" in
 
 install)
+  mvn -e install:install-file
```
romani, line 8: I do not understand this hack.
Please do:
```
git clone .....medata.git
CD metadata
Mvn install
CD ../

```

pom.xml
```diff
@@ -288,6 +288,31 @@
...
+    </dependency>
+    <dependency>
+        <groupId>org.reflections</groupId>
+        <artifactId>reflections</artifactId>
+        <version>0.9.10</version>
+    </dependency>
+    <dependency>
+      <groupId>org.example</groupId>
+      <artifactId>metadata-gen</artifactId>
+      <version>1.0</version>
```
romani, line 314: There should be snapshot version

src/main/resources/org/sonar/plugins/checkstyle/metadata/additional-rules-meta.yml
romani: "Ready" mean that Check has no ways to be customized in sonar Ui so it is ready to use by activation.

If Check in configurable, it become a template like, so user needs configure it, give it a name(instance name) and only after activate by name of instance.

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadata.java
romani: Please follow.

src/main/resources/org/sonar/plugins/checkstyle/metadata/additional-rules-meta.yml
gaurabdg: I didn't get it, every check has default values and don't need compulsory reconfiguring right? Can you please reframe the question?

src/main/resources/org/sonar/plugins/checkstyle/rule-deletedChecks.xml
gaurabdg: To test whether the rule produced by the metadata utily matches the ones produced by default by the plugin, for every new check(those which are not present in `rule-deletedChecks.xml`), I have checked their properties against the original rule properties in `testCreate` UT. I hope I could explain.

.ci/travis.sh
gaurabdg: Done

pom.xml
gaurabdg: Done

src/main/resources/org/sonar/plugins/checkstyle/metadata/additional-rules-meta.yml
romani: In sonar Ready does not mean ready to use in default form. It means that single instance of Check is enough for all and no other instances are possible/required. If Check is possible to be configurable by user, it is not ready, and user need to make Rule from it first and then activate in some profile.
In your local sonar just try to activate NoClone and AbbreviationNameCheck and will see that workflow of activation is different

rnveach:
Wercker:
````
[INFO] --- maven-install-plugin:2.4:install-file (default-cli) @ metadata-gen ---
[INFO] Installing /pipeline/source/metadata-gen/target/metadata-gen-1.0-SNAPSHOT.jar to /pipeline/cache/org/example/metadata-gen/1.0-SNAPSHOT/metadata-gen-1.0-SNAPSHOT.jar
[INFO] Installing /tmp/mvninstall8117743005364978556.pom to /pipeline/cache/org/example/metadata-gen/1.0-SNAPSHOT/metadata-gen-1.0-SNAPSHOT.pom
````

````
[ERROR] Failed to execute goal on project checkstyle-sonar-plugin: Could not resolve dependencies for project com.github.checkstyle:checkstyle-sonar-plugin:sonar-plugin:4.33-SNAPSHOT: Could not find artifact org.example:metadata-gen:jar:1.0-SNAPSHOT -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal on project checkstyle-sonar-plugin: Could not resolve dependencies for project com.github.checkstyle:checkstyle-sonar-plugin:sonar-plugin:4.33-SNAPSHOT: Could not find artifact org.example:metadata-gen:jar:1.0-SNAPSHOT
````

File: .ci/travis.sh [hunk 11]
```diff
@@ -2,6 +2,18 @@
 
 set -euo pipefail
 
+git clone https://github.com/gaurabdg/metadata-gen.git
+cd metadata-gen
+mvn clean package
+mvn install:install-file \
```
rnveach, line 8: why isn't this just `mvn install`?

rnveach: It could also be https://github.com/checkstyle/sonar-checkstyle/blob/65ba359e2a728871ba681fb9c8dc9346db6b4475/.ci/wercker.sh#L46 is overwritting the maven options so it loses track of where `.m2` is located after the install of the local.

We probably need printing on to debug the issue.

src/main/resources/org/sonar/plugins/checkstyle/rule-deletedChecks.xml
muhlba91: does this mean it's only for your proposal to know which checks you have verified and which you did not? then i'd either not commit it and keep it locally on your side or maybe movie it into a `todo` or similar directory naming it to `rules-unverified.xml`?

gaurabdg: No, it's not for that.  I wanted to test whether the new rules formed matches the old default rule properties or not. So first I created a `origRepo` with `rules.xml` and then a `modRepo` with `rule-deletedChecks.xml`. I forcefully removed 5 checks to trigger the creation of the new rules and checked the results in `testCreate` UT.
The `createMeta` creates rules for checks which are not present in the default repository formed from the xml file, thats why I created a  new file intentionally removing the checks.

.ci/travis.sh
gaurabdg: Yes the because of linking `pipeline/cache` to `root/.m2/repository`, the jar gets install in the former one, but maven still searches in the latter directory. Since, this `install` part is temporary(till PoC is approved) and will later be shipped with checkstyle, I didn't change the script for wecker, instead just added a new execution for this part in the `yml` config. It works fine now.

src/main/resources/org/sonar/plugins/checkstyle/metadata/additional-rules-meta.yml
muhlba91: for cardinality this is the new way to define it: http://javadocs.sonarsource.org/7.9/apidocs/org/sonar/api/rules/Rule.html#setIsTemplate(boolean)
(hence, if we have cardinality multiple, we need to define it as a template)

for status, i'd keep all to READY: http://javadocs.sonarsource.org/7.9/apidocs/org/sonar/api/rules/Rule.html#STATUS_BETA
it does not make sense for us to mark some as BETA or DEPRECATED, @romani ?
whatever checks we have they should all be ready and good to go unless the main project actually marks checks as deprecated, then we can add a lifetime to SQ as well: the check is always READY unless the main repo has a Deprecated annotation present, then it's DEPRECATED and will be removed at some point automatically anyways.

gaurabdg: @muhlba91 @romani @rnveach To determine cardinality I think we should follow the rules of https://github.com/checkstyle/sonar-checkstyle/issues/87#issue-214878127
Also, I think making multiple instances of a check makes sense when there are different targets to work on (token set/file extensions) right?
So, the condition to have multiple cardinality should be that the check must have either tokens/fileExtension/(or anything I am mising out), and then in addition to these some other configurable properties.

gaurabdg: After discussing with @romani it has been decided that
1. Cardinality will be always single if there are no configurable property and multiple otherwise in all cases
2. Priority will be MINOR by default
3. Status will be READY by default
4. Tag will have the check package name as tag(imports, headers, block, etc.)

So, only tag will be present as an additional detail YML config, if the user wants to specify some other tag.(@muhlba91 Is it possible for users to specify tags in sonarUI? Then we should remove it from YML, and there will be no need to maintain the YML, hence achieving full automation).

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckUtil.java
romani: you do have fully qualified name in meta files - you do not need to scan for classes.
you scan only for metafiles or old configs. get from them fully qualified name and load class by it.
please do.

src/main/resources/org/sonar/plugins/checkstyle/metadata/additional-rules-meta.yml
romani: tags should be taken from package name of this level https://github.com/checkstyle/checkstyle/tree/master/src/main/java/com/puppycrawl/tools/checkstyle/checks

if module is on this level, then we do no tag. You will see that some Checks at placed on this level for some historical reason.
It will work only for standard Checks, that is still good. But it will not work for thirparty Checks (unpredictable package anmes). So they might need to provide tags to be placed in same tags as standard Checks. Please look at sevntu Checks.
So we need support of yml but for standard Checks we can omit it, or we can provide tags for Checks that are in root package
TodoComment - https://checkstyle.org/config_misc.html#TodoComment , so we can place "misc" tag on it.
https://checkstyle.org/config_misc.html#UncommentedMain . looks like almost all Checks in this root package should have "misc" tag, it might be good to match website grouping.
So lets support this yml file and place there some overrides ... for example Indentation. it is special java package, but group in website is "misc", so lets put tag "misc" on it. All other standard Check in root package could have "misc" tag automatically. So we will have yml in plugin and there will be only "Indentation".
All thirdparty checks (sevntu, ....)  should not fall in "misc" tag if no tags provided.

gaurabdg: > and there will be only "Indentation".

I didn't understand. When specifying the IndentationCheck in yml, there will be misc tag in the config. What else are you trying to imply?

> All thirdparty checks (sevntu, ....) should not fall in "misc" tag if no tags provided.

It should fall in misc category by default right?

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckUtil.java
gaurabdg: Done

src/main/resources/org/sonar/plugins/checkstyle/metadata/additional-rules-meta.yml
romani: >  What else are you trying to imply?

nothing else. in Config we will do overrides of what you determine in from package name. By package name IndentationCheck should has "indentation" tag, but by means of this yml file it will be "misc".

> It should fall in misc category by default right?

no. If no details(no yml file) provided by extension - no tags at all. If extension would like to join standard tags it should provide mapping of Check to them in yml file or old-style config. If extension would like special tags - yml file is required.
Tags are not critical item in sonar. It is easy to find Check even without tags.

romani:
snyk can be ignored, it is not related, no vulnerabilities, but something is broken on snyk side or in our configuration.

should be ignored in this PR.

src/main/resources/org/sonar/plugins/checkstyle/metadata/additional-rules-meta.yml
gaurabdg: Okay, done.

src/main/resources/org/sonar/plugins/checkstyle/metadata/additional-rules-meta.yml
```diff
@@ -0,0 +1,4 @@
+rules:
+    - rule: EmptyBlockCheck
+    - rule: JavadocPackageCheck
+      tag: comment
```
romani, line 4: please put here Indentation only.

and name it `rules-meta.yml`, it is not additional, it is override

src/test/java/org/sonar/plugins/checkstyle/CheckstyleRulesDefinitionTest.java
```diff
@@ -75,6 +75,7 @@ public void test() {
                 assertThat(param.description()).overridingErrorMessage(
                         "Description is not set for parameter '" + param.name() + "' of rule '"
                                 + rule.key()).isNotNull();
+
```
romani, line 78: remove this change.

src/main/resources/org/sonar/plugins/checkstyle/rule-deletedChecks.xml
romani: Please move this file to test resources

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckUtil.java
```diff
@@ -0,0 +1,99 @@
...
+        }
+        else if (AbstractCheck.class.isAssignableFrom(checkResult.getClass())) {
+            final AbstractCheck check = (AbstractCheck) checkResult;
+            result = getTokenText(false, check.getAcceptableTokens(),
+                    check.getRequiredTokens());
+        }
+        return result;
+    }
+
+    public static String getTokenText(boolean isJavadocCheck, int[] tokens, int... subtractions) {
```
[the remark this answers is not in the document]
gaurabdg: the last argumnet has to be of varargs type according to a PMD violation

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckUtil.java
```diff
@@ -0,0 +1,99 @@
...
+
+    public static Object getCheck(String checkName) {
+        final ClassLoader classLoader = CheckstyleMetadata.class.getClassLoader();
+        try {
+            final Set<String> packageNames = PackageNamesLoader.getPackageNames(classLoader);
+            return new PackageObjectFactory(packageNames, classLoader)
+                    .createModule(checkName.substring(0, checkName.length() - "Check".length()));
+        }
+        catch (CheckstyleException ex) {
+            throw new IllegalStateException("exception occured during getCheck", ex);
```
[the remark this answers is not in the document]
gaurabdg: Done

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadata.java
```diff
@@ -0,0 +1,266 @@
...
+        }
+    }
+
+    private Class<?> getClass(String checkName) {
+        final ClassLoader loader = getClass().getClassLoader();
+        try {
+            return Class.forName(checkName, true, loader);
+        }
+        catch (ClassNotFoundException ex) {
+            throw new IllegalStateException("exception occured during getClass", ex);
```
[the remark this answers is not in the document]
gaurabdg: Done

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadata.java
```diff
@@ -0,0 +1,266 @@
...
+        }
+    }
+
+    private boolean isTemplateRule(String checkName) {
+        boolean result = false;
+        for (Method method : getClass(checkName).getMethods()) {
+            if (isSetter(method)) {
```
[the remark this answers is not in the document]
gaurabdg: Done

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadata.java
```diff
@@ -0,0 +1,266 @@
...
+    private static Map<String, SonarRulePropertyLoader.AdditionalRuleProperties>
+        getAdditionalDetails(String fileName) {
+        final ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
+        mapper.configure(DeserializationFeature.ACCEPT_SINGLE_VALUE_AS_ARRAY, true);
+        final SonarRulePropertyLoader sonarRulePropertyLoader;
+        try (InputStream inputStream = CheckstyleMetadata.class.getResourceAsStream(fileName)) {
+            sonarRulePropertyLoader = mapper.readValue(inputStream, SonarRulePropertyLoader.class);
+        }
+        catch (IOException ex) {
+            throw new IllegalStateException("Exception during yaml loading", ex);
```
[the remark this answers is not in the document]
gaurabdg: Done

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadata.java
```diff
@@ -0,0 +1,266 @@
...
+        ModuleDetails moduleDetails = null;
+        try {
+            final InputStream inputStream = getClass().getResourceAsStream(checkName + ".xml");
+            if (inputStream != null) {
+                moduleDetails = new XMLReader().read(inputStream, ModuleType.CHECK);
+                inputStream.close();
+            }
+        }
+        catch (IOException ex) {
+            throw new IllegalStateException("exception occured in loadMeta", ex);
```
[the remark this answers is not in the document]
gaurabdg: Done

src/main/resources/org/sonar/plugins/checkstyle/metadata/additional-rules-meta.yml
gaurabdg: Done

src/test/java/org/sonar/plugins/checkstyle/CheckstyleRulesDefinitionTest.java
gaurabdg: Done

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadata.java
```diff
@@ -0,0 +1,261 @@
...
+        }
+        else {
+            result = configData.getTag();
+        }
+        return result;
+    }
+
+    private static Map<String, SonarRulePropertyLoader.AdditionalRuleProperties>
+        getAdditionalDetails(String fileName) {
+        final ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
```
romani, line 194: please add empty line above this line, to claerly show where method body is started.

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckUtil.java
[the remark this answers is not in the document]
romani: ok, lets keep it for now.

pom.xml
```diff
@@ -288,6 +288,31 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+        <groupId>com.fasterxml.jackson.dataformat</groupId>
+        <artifactId>jackson-dataformat-yaml</artifactId>
+        <version>2.3.0</version>
```
romani, line 294: please use latest version https://mvnrepository.com/artifact/com.fasterxml.jackson.dataformat/jackson-dataformat-yaml

to fix snyk violations:
image

please do the same for all other dependencies.

romani:
@gaurabdg , Why I still see no tags in screenshots ?

src/main/java/org/sonar/plugins/checkstyle/CheckstyleRulesDefinition.java
```diff
@@ -99,4 +116,5 @@ private static void loadNames(NewRepository repository, InputStream stream) {
             });
         }
     }
+
```
romani, line 119: remove

src/test/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadataTest.java
```diff
@@ -0,0 +1,152 @@
...
+        });
+    }
+
+    private ModuleDetails loadMeta(String checkName) {
+        final ModuleDetails moduleDetails;
+        try (InputStream inputStream = getClass().getResourceAsStream(checkName + ".xml")) {
+            moduleDetails = new XMLReader().read(inputStream, ModuleType.CHECK);
+        }
+        catch (IOException ex) {
+            throw new IllegalStateException("exception occured during loadMeta", ex);
```
romani, line 125: checkname to print.

src/main/java/org/sonar/plugins/checkstyle/CheckstyleRulesDefinition.java
gaurabdg: Done

src/test/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadataTest.java
gaurabdg: Done

gaurabdg:
> @gaurabdg , Why I still see no tags in screenshots ?

Sorry, I missed it. I have uploaded in the PR description, please have a look.

Issue #335, opened 2020-09-19
Bad name of modules in metadata for JavaNCSS and NPathComplexity


romani:
Bad names of Checks in metadata files:
![image](https://user-images.githubusercontent.com/812984/93668807-f47ff200-fa43-11ea-951b-679b71c9670a.png)

Expected names same as in html doc:
Java NCSS
NPath Complexity

https://checkstyle.org/config_metrics.html#NPathComplexity
https://checkstyle.org/config_metrics.html#JavaNCSS

name is metadata is good
https://github.com/checkstyle/checkstyle/blob/764805d78fe676bb15bae46e3be227c9084a0b99/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/metrics/NPathComplexityCheck.xml#L5
we need to update logic to process such Checks in special way (hardcode name in plugin or do not split NPath and NCSS).
Main library can not change name of module as it will be big breaking compatibility, but it is unlikely happen in future as chekcstyle library enforced name format to prevent such problems in future.

romani:
@gaurabdg, thanks lot

romani:
Fix is merged

PR #336, opened 2020-09-20
Issue #335: Bad name of modules in metadata for JavaNCSS and NPathCom…

gaurabdg:
#335

### DIFF
--- a/src/main/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadata.java
+++ b/src/main/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadata.java
@@ -39,14 +39,13 @@ import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.puppycrawl.tools.checkstyle.meta.ModuleDetails;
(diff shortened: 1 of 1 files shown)


src/main/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadata.java
```diff
@@ -314,10 +314,18 @@ private static String getRuleTag(String checkPackage,
     public static String getFullCheckName(String checkName) {
         final int capacity = 1024;
         final StringBuilder result = new StringBuilder(capacity);
-        for (int i = 0; i < checkName.length(); i++) {
-            result.append(checkName.charAt(i));
-            if (i + 1 < checkName.length() && Character.isUpperCase(checkName.charAt(i + 1))) {
-                result.append(' ');
+        if ("JavaNCSS".equals(checkName)) {
```
muhlba91, line 317: can we move this to a map:
JavaNCSS -> Java NCSS
NPathComplexity -> NPath Complexity

if we find the key in the map, append the value, otherwise get the name using the for loop.

this allows us to very easily add other such cases with kinda 1 line of code. :)

muhlba91 (changes requested):

romani (approved):
As map is used , ok to merge

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadata.java
gaurabdg: @muhlba91 Done.

romani (approved):

Issue #331, opened 2020-09-12
upgrade to checkstyle 8.36.1


romani:
https://checkstyle.org/releasenotes.html#Release_8.36.1

muhlba91:
@romani @gaurabdg when installing the new version on SQ, I see the following deviation.

CS 8.35:
<img width="1010" alt="Screenshot 2020-09-13 at 1 37 57 PM" src="https://user-images.githubusercontent.com/653739/93017145-af346e00-f5c6-11ea-84b5-d644d02efa23.png">

CS 8.36.1:
<img width="1032" alt="Screenshot 2020-09-13 at 1 39 35 PM" src="https://user-images.githubusercontent.com/653739/93017153-b9566c80-f5c6-11ea-9b85-808f0f2a2f35.png">

Almost all rules/checks (except for `SuppressWarningsFilter`) are template rules now.

@romani is this an expected change from CS side?
I understand it logically that I might enable different variations of checks when I have at least 1 parameter which is non-enum like. Examples would be `AnnotationLocationCheck` or `AvoidStarImportCheck`.
However, for others (e.g. `AvoidInlineConditionalsCheck`) they do not even have one parameter and, again others (e.g. `AnnotationOnSameLineCheck`) have only the tokens which is an enum list.
There are also other checks with exactly one parameter (true/false like `AvoidNestedBlocksCheck`) where this doesn't make sense to me.

Anyhow, if this is intended on CS side, I'll perform the release nevertheless.
Otherwise, @gaurabdg please investigate, create issue and fix. :)

romani:
https://checkstyle.org/config_coding.html#AvoidInlineConditionals has no properties, it should not be template

https://checkstyle.org/config_annotation.html#AnnotationOnSameLine is not obvious. From first sight it not reasonable to make it template as only tokens are modifiable, but user might create two instances of Check to use different suppressions for them or use different message on violations.

muhlba91:
i suspect this check resulting in wrong results: https://github.com/checkstyle/sonar-checkstyle/blob/master/src/main/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadata.java#L145

`getMethods()` also retrieves inherited methods to which e.g. `setTokens(...)` belongs. as all checks are inheriting from the base class `AbstractCheck` all of them will have a setter, and become a template.

`AvoidInlineConditionalsCheck` also inherits from this class, and becomes a template , although it doesn't use tokens, i presume?

[...]

PR #338, opened 2020-09-26
Issue #331: add template backward compatibility

gaurabdg:
#331

### DIFF
--- a/src/main/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadata.java
+++ b/src/main/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadata.java
@@ -100,9 +100,10 @@ public class CheckstyleMetadata {
                     final SonarRulePropertyLoader.AdditionalRuleProperties additionalDetails =
                             additionalRuleData.get(check.getKey());
                     final RulesDefinition.NewRule rule =
-                            repository.createRule(moduleDetails.getFullQualifiedName());
+                            repository.createRule(moduleDetails.getFullQualifiedName()
+                                    + "template");
                     rule.setHtmlDescription(moduleDetails.getDescription())
-                            .setName(getFullCheckName(moduleDetails.getName()))
+                            .setName(getFullCheckName(moduleDetails.getName()) + " Template")
                             .setInternalKey(getInternalKey(moduleDetails))
                             .setSeverity("MINOR")
                             .setStatus(RuleStatus.READY);
--- a/src/test/java/org/sonar/plugins/checkstyle/CheckstyleRulesDefinitionTest.java
+++ b/src/test/java/org/sonar/plugins/checkstyle/CheckstyleRulesDefinitionTest.java
@@ -21,7 +21,10 @@ package org.sonar.plugins.checkstyle;
 
 import static org.fest.assertions.Assertions.assertThat;
 
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
 
 import org.junit.Rule;
 import org.junit.Test;
@@ -61,7 +64,23 @@ public class CheckstyleRulesDefinitionTest {
         assertThat(repository.language()).isEqualTo("java");
 
         final List<RulesDefinition.Rule> rules = repository.rules();
-        assertThat(rules).hasSize(179);
+        final Map<String, Integer> ruleCounts = new HashMap<>();
+        rules.forEach(rule -> {
+            final String name = rule.key().replace("template", "");
+            ruleCounts.merge(name, 1, Integer::sum);
+        });
+        final List<String> duplicatedRuleWithTemplate = ruleCounts.entrySet().stream()
+                .filter(entry -> entry.getValue() > 1)
+                .map(entry -> entry.getKey())
+                .collect(Collectors.toList());
+        final List<String> rulesWithDuplicateTemplate = ruleCounts.entrySet().stream()
+                .filter(entry -> entry.getValue() == 1)
+                .map(entry -> entry.getKey())
+                .collect(Collectors.toList());
+        // such number should not change during checkstyle version upgrade
+        assertThat(duplicatedRuleWithTemplate).hasSize(174);
+        // all new Rules should fall in this group
+        assertThat(rulesWithDuplicateTemplate).hasSize(5);
 
         for (RulesDefinition.Rule rule : rules) {
             assertThat(rule.key()).isNotNull();
--- a/config/suppressions.xml
+++ b/config/suppressions.xml
@@ -32,5 +32,5 @@
     <suppress checks="MagicNumber" files=".*[\\/]src[\\/]test[\\/]"/>
     <suppress checks="AnonInnerLength" files=".*[\\/]src[\\/]test[\\/]"/>
     <suppress checks="ClassDataAbstractionCoupling" files=".*[\\/]src[\\/]test[\\/]"/>
-    <suppress checks="TrailingComment" files=".*[\\/]CheckstyleMetadata\.java"/>
+    <suppress checks="TrailingComment" files=".*[\\/]Checkstyle(RulesDefinition|Metadata)\.java"/>
 </suppressions>
--- a/pom.xml
+++ b/pom.xml
@@ -576,6 +576,7 @@
                     <exclude>org.sonar.plugins.checkstyle.CheckstyleSeverityUtils</exclude>
                     <exclude>org.sonar.plugins.checkstyle.CheckstyleConfiguration</exclude>
                     <exclude>org.sonar.plugins.checkstyle.CheckstyleProfileImporter</exclude>
+                    <exclude>org.sonar.plugins.checkstyle.CheckstyleRulesDefinition</exclude>
                     <exclude>org.sonar.plugins.checkstyle.metadata.CheckstyleMetadata</exclude>
                     <exclude>org.sonar.plugins.checkstyle.metadata.CheckUtil</exclude>
                   </excludes>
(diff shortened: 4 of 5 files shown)


romani:
@gaurabdg, please share screenshots on how it looks like. Did you test backwards compatibility ?

CI is not happy.

gaurabdg:
> @gaurabdg, please share screenshots on how it looks like.
Screenshot 2020-09-27 at 2 58 34 AM

> Did you test backwards compatibility ?

Yes, created config and ran `mvn sonar:sonar`.

> CI (sonar) is not happy.

@muhlba91 Please look into this, I just reintroduced old code, which are giving errors.

muhlba91:
@gaurabdg we had similar problems already and copied over code from this library. please copy-paste the methods used into sonar-checkstyle in the same style like we did with https://github.com/checkstyle/sonar-checkstyle/blob/4.33/src/main/java/org/sonar/plugins/checkstyle/CheckstyleRulesDefinition.java#L73 (`CheckstyleRulesDefinition::loadNames`).

gaurabdg:
@muhlba91 I dont think you are understanding the problem, the code(which existed in master before I removed) used is depreacted.
I am adding `SuppressWarning` annotation to pass sonar CI.

src/test/java/org/sonar/plugins/checkstyle/CheckstyleRulesDefinitionTest.java
```diff
@@ -61,7 +61,7 @@ public void test() {
         assertThat(repository.language()).isEqualTo("java");
 
         final List<RulesDefinition.Rule> rules = repository.rules();
-        assertThat(rules).hasSize(179);
+        assertThat(rules).hasSize(353);
```
romani, line 64: this should be 179 * 2 = 358 - (amount of Checks that are new)

Looks like we have 5 new Checks, but I see only 4 new checks https://checkstyle.org/releasenotes.html 

IllegalIdentifierName
RecordComponentNumberCheck
RecordTypeParameterNameCheck
PatternVariableName.

https://checkstyle.org/releasenotes.html#Release_8.33 we introduced NoCodeInFile and it is old metadata, so most likely has to be duplicated.

Please help to explain why `353`.

muhlba91:
@gaurabdg yes, the code in sslr bridge is deprecated and they will remove it at some point. since we rely on that code part, we should move it to our code base and not rely on sslr bridge to provide it. once it‘s in our code base, we don‘t have to deprecate it.

problem will be: in a future release they will just delete the code. we still need it, so we should copy the code from sslr bridge to our code base.
that was the same with loadNames. they simply removed it (as far as i still remember) and we had to copy it. i‘d rather not like running into the same situation if i can avoid it already...

however, i also see this not as part of this particular issue to solve. if there‘s time for you, please copy it over, otherwise, i‘ll perform the transition in a future commit. :) 

good to go from my side once the check count question is resolved.

romani:
Lets postpone deprecation nuance for later. We need to make it work first.

@muhlba91 , please confirm one more time that your existing sonar config is migrated correctly to new version of sonar.

romani:
Travis CI faililng with:
```
[INFO] --- maven-checkstyle-plugin:3.1.1:check (checkstyle-check) @ checkstyle-sonar-plugin ---
[INFO] There is 1 error reported by Checkstyle 8.35 with https://raw.githubusercontent.com/checkstyle/checkstyle/checkstyle-8.35/config/checkstyle_checks.xml ruleset.
[ERROR] src/main/java/org/sonar/plugins/checkstyle/CheckstyleRulesDefinition.java:[60,23]
 (annotation) SuppressWarnings: The warning 'squid:S1874' cannot be suppressed at this location.
            Only few javac warnings are allowed to suppress.            
If try to suppress checkstyle/pmd/..... violation please do this in their config file.            
If you try to suppress IntelliJ IDEA inspection,            
please use javadoc block tag @noinspection
``` 

I am ok to suppress this violation for now in suppression.xml .

romani:
@gaurabdg , please reply.

src/test/java/org/sonar/plugins/checkstyle/CheckstyleRulesDefinitionTest.java
romani: I found missed by me Check - https://checkstyle.org/config_javadoc.html#RequireEmptyLineBeforeBlockTagGroup
there was mistake in release notes, bad issue title.

fixed.

romani: I updated code to validate counts more clearly

romani (approved):
@muhlba91 , please finalize review and testing.

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadata.java
```diff
@@ -100,9 +100,10 @@ public void createRulesWithMetadata() {
                     final SonarRulePropertyLoader.AdditionalRuleProperties additionalDetails =
                             additionalRuleData.get(check.getKey());
                     final RulesDefinition.NewRule rule =
-                            repository.createRule(moduleDetails.getFullQualifiedName());
+                            repository.createRule(moduleDetails.getFullQualifiedName()
+                                    + "template");
```
romani, line 104: the only minor item it have for this line.
it make rule id looks like 
`com.puppycrawl.tools.checkstyle.checks.naming.AbbreviationAsWordInNameChecktemplate`
but probably better to do
`com.puppycrawl.tools.checkstyle.checks.naming.AbbreviationAsWordInNameCheckTemplate`

@muhlba91 , please share your thoughts on this.

muhlba91:
@romani @gaurabdg i was not able to simply run sonar after updating:
```
Caused by: java.lang.IllegalArgumentException: given name empty
        at com.puppycrawl.tools.checkstyle.utils.TokenUtil.getTokenId(TokenUtil.java:174)
        at com.puppycrawl.tools.checkstyle.TreeWalker.registerCheck(TreeWalker.java:212)
        at com.puppycrawl.tools.checkstyle.TreeWalker.setupChild(TreeWalker.java:130)
        at com.puppycrawl.tools.checkstyle.api.AutomaticBean.configure(AutomaticBean.java:201)
        at com.puppycrawl.tools.checkstyle.Checker.setupChild(Checker.java:477)
        at com.puppycrawl.tools.checkstyle.api.AutomaticBean.configure(AutomaticBean.java:201)
        at org.sonar.plugins.checkstyle.CheckstyleExecutor.executeWithClassLoader(CheckstyleExecutor.java:102)
        ... 217 more
```

what i did is this: start SQ with 8.35 (so the previous version), bulk enable all checkstyle rules (no templates though), run analysis, update to 8.36-SNAPSHOT, restart SQ, run analysis.

i did not receive any further output from `gradle sonar --stacktrace` but i'll add the generated checkstyle.xml. @romani could you please have a quick look at it?

other than that, if we cannot figure out a good way we must decide on either a major version release to indicate incompatibility or mix and match forever but with limiting a user's functionality and also having to maintain those old checks manually. any suggestions on this one?
personally, i do not see a problem in a major version upgrade and breaking compatibility and i probably think it will be inevitable in the future.

[checkstyle.txt](https://github.com/checkstyle/sonar-checkstyle/files/5348270/checkstyle.txt)

romani:
@gaurabdg, please help to figure out what is problem.

rnveach:
The config with the issue is:
````
		<module name="DescendantToken">
			<property name="severity" value="info" />
			<property name="maximumDepth" value="2147483647" />
			<property name="maximumNumber" value="2147483647" />
			<property name="tokens" value="empty" />
			<property name="minimumNumber" value="0" />
			<property name="minimumDepth" value="0" />
			<property name="sumTokenCounts" value="false" />
		</module>
````

Notice the line `<property name="tokens" value="empty" />`.

I assume this relates back to https://github.com/checkstyle/checkstyle/blob/39b02e16d55a081ba779c0bdb2318ab30eaea515/src/main/resources/com/puppycrawl/tools/checkstyle/meta/checks/DescendantTokenCheck.xml#L49

romani:
@gaurabdg, please help to fix metadata file

strkkk:
@romani I can suggest following. In DescendantTokenCheck value for tokens property will be set to empty string (or any value that should be treated as empty).
Then, add to setTokens method filtering for empty strings
https://github.com/checkstyle/checkstyle/blob/58bae0f3defcf7ba4a9fa69bed97aadcb35eb2ae/src/main/java/com/puppycrawl/tools/checkstyle/api/AbstractCheck.java#L95-L97

By doing that, this line in registerCheck method
https://github.com/checkstyle/checkstyle/blob/58bae0f3defcf7ba4a9fa69bed97aadcb35eb2ae/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java#L201

will return empty set of tokens, which is correct.

It seems to me like a very small fix with minimal regression possible.

romani:
I tested suggested by me fix in meta see PR: https://github.com/checkstyle/checkstyle/pull/8917

Screenshot from 2020-10-22 07-22-02
Screenshot from 2020-10-22 07-22-45

romani:
@muhlba91 , please build checkstyle from my PR `mvn install -Pno-validations` and use it this branch.
Your extra help with testing will help a lot. If any problems found I can fix them before next checkstyle release (Oct 31).

muhlba91:
fyi: i‘ll run my tests most likely tomorrow evening. unfortunately, those weeks are busier on my side.

romani:
Thanks a lot.
I will release checkstyle library with required for this PR update update on this weekend.

muhlba91:
tested and looking good from my side:
- 8.35: enable all rules & run analysis
- update CS plugin & run analysis
- enable all rules of snapshot & run analysis
- enable some template rules & run analysis

romani:
@muhlba91 , I released https://checkstyle.org/releasenotes.html#Release_8.37 , looks like we need to merge this PR
and upgrade to 8.37 and do final test.

PR #355, opened 2021-02-16
Issue #353: update to CS 8.40

### LINKED ISSUE #353: upgrade to checkstyle 8.40
https://checkstyle.org/releasenotes.html#Release_8.40

muhlba91:
fix #353.

New check:
Screenshot 2021-02-16 at 7 57 21 PM

DesignForExtension new property (default value):
Screenshot 2021-02-16 at 7 57 31 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ Compatibility matrix from checkstyle team:
 
 Checkstyle Plugin|Sonar min|Sonar max|Checkstyle|Jdk
 -----------------|---------|---------|----------|---
+8.40|7.9  |7.9+|8.40|1.8
 8.39|7.9  |7.9+|8.39|1.8
 8.38|7.9  |7.9+|8.38|1.8
 8.37|7.9  |7.9+|8.37|1.8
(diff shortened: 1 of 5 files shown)


romani (approved):
Why we need plugin personal metadata update ?
Whole idea of summer project was to avoid this. Where we lost something ?

muhlba91:
We apply the new metadata only to new checks due to backwards compatibility.

Also, this update affects the non-template rule version of this check but the template rule version of it gets auto-generated by metadata.

romani:
@muhlba91 , it would be awesome to think on how declare to users breaking compatibility time/version and remove all non-template metadata, or find a way for user on how to migrate configurations.
Users will not like it but it will reduce maintenance requirements for project so project (this plugin) will continue to be supported.
if project is easy support it will be alive and we will always be in use. if we will always think about users only to reduce their pain, we will end up in pain for maintainers, so there might be no new versions and project become abandoned so users will not get new versions also. Eventually they will find time to migrate, BUT it is significantly harder to find maintainer for project.

PR #366, opened 2021-04-20
WIP: Issue #361: update to CS 8.41.1

### LINKED ISSUE #361: upgrade to checkstyle 8.41.1
https://checkstyle.org/releasenotes.html#Release_8.41.1

muhlba91:
fix #361

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -93,7 +93,7 @@
   </ciManagement>
 
   <properties>
-    <checkstyle.version>8.40</checkstyle.version>
+    <checkstyle.version>8.41.1</checkstyle.version>
     <sonar.version>7.9</sonar.version>
     <sonar-java.version>6.0.0.20538</sonar-java.version>
     <maven.checkstyle.plugin.version>3.1.1</maven.checkstyle.plugin.version>
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ Compatibility matrix from checkstyle team:
 
 Checkstyle Plugin|Sonar min|Sonar max|Checkstyle|Jdk
 -----------------|---------|---------|----------|---
+8.41|7.9  |7.9+|8.41.1|1.8
 8.40|7.9  |7.9+|8.40|1.8
 8.39|7.9  |7.9+|8.39|1.8
 8.38|7.9  |7.9+|8.38|1.8


muhlba91:
@romani can you help here?

the error received does not (seem to) relate to the plugin, as far as i figured out.

research shows that jdk bundled the xml reader stuff at some point but some plugins still seem to use xerces. depending on the implementation used for DocumentBuilderFactory it can be that xerces results in such errors. i am more puzzled by the fact that this error never occurred before.

either way, dependency tree analysis shows excluding xerces coming in from direct dependencies of the plugin does not resolve it.
there's a sonar java dependency on xerces; however, we have that exact version of sonar java since upgrading to 7.8 as the main sonar version and never had problems (neither with the previous release)  i also do not find any release version not including the dependencies containing xerces; hence, i doubt this dependency is the cause...

digging deeper in the stacktrace this line could (?) be the culprit:
https://github.com/checkstyle/checkstyle/blob/checkstyle-8.41.1/src/main/java/com/puppycrawl/tools/checkstyle/meta/XmlMetaReader.java#L127
resorting to `DocumentBuilderFactory docBuilderFactory = DocumentBuilderFactory.newInstance("com.sun.org.apache.xerces.internal.jaxp.DocumentBuilderFactoryImpl", this.getClass().getClassLoader());` as suggested by some as a workaround and using an `internal` class does not really satisfy...

also just to be sure, reverting the last guava bump change does not resolve the issue which would be the only change happening in this repo since our last release.

i wonder whether you changed anything about your build environment and/or dependencies in the main CS project during the 8.41(.1) release and/or ran into this issue?

see:
- https://stackoverflow.com/questions/25453042/how-to-disable-accessexternaldtd-and-entityexpansionlimit-warnings-with-logback
- https://docs.oracle.com/javase/tutorial/jaxp/properties/error.html
- https://stackoverflow.com/questions/53299280/java-and-xerces-cant-find-property-xmlconstants-access-external-dtd

muhlba91:
@romani ping. any ideas on this issue?

romani:
I keep your problem in notification list as not replied, I think I will be able to help only in September, after Gsoc completion.

pom.xml
```diff
@@ -93,7 +93,7 @@
   </ciManagement>
 
   <properties>
-    <checkstyle.version>8.40</checkstyle.version>
+    <checkstyle.version>8.41.1</checkstyle.version>
```
muhlba91, line 96: this should not be changed. when performing a release, we run mvn release which increments the version number automatically.

romani:
Closed in favor of https://github.com/checkstyle/sonar-checkstyle/pull/380

PR #381, opened 2021-10-17
infra: move all travis jobs to GitHub workflows

nmancus1:
After discussion with @rnveach , we have decided to move travis CI tasks to github workflows. I have left `travis.yml` in place for now; I am not sure if we should delete it or just disable travis on this repo (or both).

Edit: Travis is currently not running


muhlba91:
no, i do not have permissions to change the repo settings.

PR #387, opened 2021-10-29
Issue #384: update to sonarqube 8.9 LTS

### LINKED ISSUE #384: SonarQube LTS 8.9
SonarQube's new LTS version is 8.9. We should facilitate an upgrade to use this as our compatibility target.

SQ will disable listing plugins with a lower bound than 8.9 soon. 8.42 is on the marketplace with this lower bound already; however, we should upgrade to be on-par with SQ LTS.

muhlba91:
fix #384 

plugin starting and working (tested template and non-template check):
Screenshot 2021-10-29 at 8 44 52 AM

### DIFF
--- a/.ci/ci.sh
+++ b/.ci/ci.sh
@@ -9,13 +9,14 @@ install)
   ;;
 
 integration-tests)
-  SONAR_APP_VERSION="7.9.2"
+  SONAR_APP_VERSION="8.9.3.48735"
   if [[ ! -f ~/.m2/sonar-application-$SONAR_APP_VERSION.zip ]]; then
     URL="https://repox.jfrog.io/repox/sonarsource/org/sonarsource/sonarqube/"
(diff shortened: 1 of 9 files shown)


romani:
@muhlba91 , please update https://github.com/checkstyle/sonar-checkstyle/wiki/How-to-build-and-test-plugin-on-local
to new version.
I use this page a lot, when I come back to plugin, I do not remember all that nuances.

romani (approved):
ok to merge

romani:
@muhlba91 , does it make sense to combine this update with upgrade to checkstyle 9.0 ?
so it will be to major upgrades in same time.

muhlba91:
@romani changed the wiki page and, yes, i would like to combine both updates with the 9.0 release.
it fits well and i can anyways only submit a plugin with the lower boundary of 8.9 (8.42 and 8.45.1 is in the marketplace like this) as SQ will drop all plugins lower than than min. bound from the marketplace soon.

romani:
Keep in mind that marketplace is good, but  a lot user use old version of Sonar and plugins, manually installed

PR #433, opened 2022-10-13
Issue #431: upgrade checkstyle to 10.3.3

muhlba91:
fixes #431
updates CS to 10.3.3

PR #434, opened 2022-10-18
infra: fix build and plugin execution

rnveach:
Identified at https://github.com/checkstyle/sonar-checkstyle/pull/433#issuecomment-1281704317

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,9 +1,20 @@
 version: 2
 jobs:
-  build:
+  build-project:
     docker:
-      - image: checkstyle/sonarqube-maven-git:7.7-community-8c1737e
+      - image: cimg/openjdk:11.0.16
 
+    steps:
+      - checkout
+      - run: mvn package -Pno-validations
+      - persist_to_workspace:
+          root: /home/circleci
+          paths:
+            - project
+
+  execute-sonar:
+    docker:
+      - image: sonarqube:9.0-community
     working_directory: ~/repo
 
     environment:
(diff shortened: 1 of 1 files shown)


rnveach:
https://app.circleci.com/pipelines/github/checkstyle/sonar-checkstyle/690/workflows/5c9a0e73-fd6f-4fa8-9a60-8384b0cbfdcf/jobs/664?invite=true#step-103-245
8 still failed.
https://github.com/SonarSource/docker-sonarqube/blob/8c80d5d9a234cba94b8472c2fee38e6229fed936/8/community/Dockerfile#L3
Docker file says it has 11, and has had it for a while.

rnveach:
Found some docker image maintenance at https://hub.docker.com/repository/docker/checkstyle/sonarqube-maven-git

rnveach:
https://app.circleci.com/pipelines/github/checkstyle/sonar-checkstyle/693/workflows/60b90dc0-263b-4a11-8013-459a9e470d60/jobs/667?invite=true#step-103-382
`7.9-community-5302c42` is failing on compiling which is different.
> No compiler is provided in this environment. Perhaps you are running on a JRE rather than a JDK?

muhlba91:
prerequisite: we will need to update the used image to use SQ 9+ as this is the current version necessary by the plugin. i did not find such image on dockerhub, though.

nrmancuso:
Build and execution of sonar takes less than 2 minutes, I do not see necessity of cache usage.

.circleci/config.yml
```diff
@@ -1,54 +1,61 @@
...
-      - image: checkstyle/sonarqube-maven-git:7.7-community-8c1737e
+      - image: cimg/openjdk:11.0.16
 
+    steps:
+      - checkout
+      - run: mvn package -Pno-validations
+      - persist_to_workspace:
+          root: /home/circleci
+          paths:
+            - project
```
nrmancuso, line 13: We do not need composed images, since we can build jar and generate class files in previous step, and persist the project directory for sonar execution job.

.circleci/config.yml
```diff
@@ -1,54 +1,61 @@
...
       # Customize the JVM maximum heap limit
       MAVEN_OPTS: -Xmx3200m
 
     steps:
-      - checkout
+      - attach_workspace:
+          at: /root
+      - run:
+          name: Install packages
+          command: apk add maven
```
nrmancuso, line 29: Installation of maven takes less than 5 seconds, I do not think we need sonar image with maven pre-installed

romani (approved):
Ok to merge.

Agree, We do not need custom image in 2022 year, docker in circleci is quick and sonar provide docker itself

rnveach:
@romani 
https://github.com/checkstyle/sonar-checkstyle/pull/439#issuecomment-1286247435
> How we can prove that plugin continue to work?

This circleci run actually runs our plugin with sonar, so this is one way to test. If we could find a way, we could configure this for integration tests.

rnveach:
@nrmancuso Thank you so much for this.

Issue #40, opened 2016-12-18
get rid of dependency to sonar-java-plugin


romani:
this dependency is required only do smth with classloader.
We should implement this ourself.
We should not depend on this artifact at all.

romani:
@muhlba91,

>Unfortunately, I don't get exactly what important pieces it extracts and I quickly skimmed through it's code as well as JavaClasspath which it utilizes to see how we could easily duplicate this ode

Checkstyle should not rely on class path. There were few Checks that tried usage of class path, but such implementation endup very problatic and unstable, so we removed it. But looks like plugin tried to address this problem of classpath, BUT it is not supported feature of checkstyle.

There are few Checks still use classpath, but they will work fine without it, as they do in other plugins.

Usage of such plugin might be a reason why users report no action at non Java files #207 . Checkstyle plugin should not use classpath and should be able to process non-java files.

romani:
@muhlba91 ,

checkstyle completely disabled any class path activity in all checks. There still such methods in Checker/Treewalker but they do nothing in core library and only exists for compatibility for some time. Will be completely removed ones maven checkstyle plugin is released after 3.1.0 .

Sonar plugin completely ok to drop any features related to classpath and remove this dependency to plugin.

romani:
Fix is merged

PR #439, opened 2022-10-20
Issue #40: removed sonar-java-plugin

rnveach:
Issue #40

Removed the dependency.

I was not seeing any classpath shenanigans mentioned in the issue unless its happening via reflection or such.

The code before this was causing issues with Eclipse:
> The package javax.xml.stream is accessible from more than one module: <unnamed>, java.xml	CheckstyleProfileImporter.java	/sonar-checkstyle/src/main/java/org/sonar/plugins/checkstyle	line 28	Java Problem

https://stackoverflow.com/questions/55571046/eclipse-is-confused-by-imports-accessible-from-more-than-one-module

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -111,19 +111,11 @@
   </properties>
 
   <dependencies>
-    <!-- till https://github.com/checkstyle/sonar-checkstyle/issues/40 -->
     <dependency>
-      <groupId>org.sonarsource.java</groupId>
-      <artifactId>sonar-java-plugin</artifactId>
-      <version>${sonar-java.version}</version>
+      <groupId>com.fasterxml.staxmate</groupId>
+      <artifactId>staxmate</artifactId>
+      <version>2.4.0</version>
       <scope>provided</scope>
-      <!-- to avoid 4.5.1 api pickup -->
-      <exclusions>
-        <exclusion>
-          <groupId>org.codehaus.sonar</groupId>
-          <artifactId>sonar-plugin-api</artifactId>
-        </exclusion>
-      </exclusions>
     </dependency>
     <!-- required to load external descriptiona, see CheckstyleRulesDefinition -->
     <dependency>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -111,19 +111,11 @@
   </properties>
 
   <dependencies>
-    <!-- till https://github.com/checkstyle/sonar-checkstyle/issues/40 -->
     <dependency>
-      <groupId>org.sonarsource.java</groupId>
-      <artifactId>sonar-java-plugin</artifactId>
-      <version>${sonar-java.version}</version>
+      <groupId>com.fasterxml.staxmate</groupId>
+      <artifactId>staxmate</artifactId>
```
rnveach, line 116: New dependency was required since the other included it and we still do as well.

src/main/java/org/sonar/plugins/checkstyle/CheckstyleSensor.java
```diff
@@ -34,7 +33,7 @@ public CheckstyleSensor(CheckstyleExecutor executor) {
 
     @Override
     public void describe(SensorDescriptor descriptor) {
-        descriptor.onlyOnLanguage(Java.KEY).name("CheckstyleSensor");
+        descriptor.onlyOnLanguage("java").name("CheckstyleSensor");
```
rnveach, line 36: This comes from `sonar-java-plugin` and so we have to hardcode it since we want to remove that dependency.

rnveach:
````
public class Java extends AbstractLanguage {

  /**
   * Java key
   */
  public static final String KEY = "java";
````

romani:
How we can prove that plugin continue to work? compilation is not enough.

muhlba91 (approved):
lgtm

rnveach:
@romani I do not know how to test this.
https://github.com/checkstyle/sonar-checkstyle/wiki/How-to-build-and-test-plugin-on-local
Does not say how to set up docker and I have never worked with it.

romani:
@muhlba91 usually runs it during release, so I hope we will do this that time also so it will be tested

muhlba91:
i usually test it e2e when opening a MR for a new version upgrade, and if other changes happen between that merge and the release i also test it again before releasing it.

i‘ll run a deeper regression test than usually after the changes being merged now.

fyi: next release will take a few more days (probably ~1.5 weeks) as i‘m on business travels since a few weeks and quite busy. i‘ll then shortly after perform the next version upgrade.

Issue #465, opened 2023-02-05
Exessive Class Loading


rnveach:
Sonar has a utility method that loads an AbstractCheck by [scanning the packages through `PackageNamesLoader` and then scanning the classloader for all modules by calling `PackageNamesLoader`](https://github.com/checkstyle/sonar-checkstyle/blob/6e3a90fdff4d8e853b0ae9305c6576925ab1842f/src/main/java/org/sonar/plugins/checkstyle/metadata/CheckUtil.java#L62-L72).

By itself, this code is fine, but it turns out this is called in [`CheckstyleMetadata`](https://github.com/checkstyle/sonar-checkstyle/blob/c9d4f493528df2fc3db1c966e2fe496f59b20fff/src/main/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadata.java#L210) for [every single module it finds](https://github.com/checkstyle/sonar-checkstyle/blob/c9d4f493528df2fc3db1c966e2fe496f59b20fff/src/main/java/org/sonar/plugins/checkstyle/metadata/CheckstyleMetadata.java#L143).

`PackageNamesLoader.getPackageNames` and `PackageObjectFactory` have no static caching in their implementations. `PackageObjectFactory` only has some instance caching.

So in sum, each module it finds it is constantly rescanning the entire classpath just to load that one single module. Checkstyle has hundreds of modules, so this is hundreds of rescans. This is a lot of wasted resources.

We need to redo this code.

rnveach:
Creating a dummy test with a lot of NO-OPs, `new CheckstyleMetadata(repository).createRulesWithMetadata();` takes 1.1 seconds to complete with a total 47 calls to `CheckUtil.getCheck`. With same basic caching it did reduce to 1 second.

All the modules being loaded are Checkstyle's in the test, so it doesn't go through any 3rd party scanning. 3rd party modules like sevntu will most likely be hit harder.

It is still a waste to constantly re-run `PackageNamesLoader`.

rnveach:
Fix was merged

PR #466, opened 2023-02-05
Issue #465: adds caching for metadata around PackageObjectFactory

rnveach:
Issue #465

I did a no harm test by having a test run `new CheckstyleMetadata(repository).createRulesWithMetadata();`. It loaded 40 checks with no issues.

### DIFF
--- a/src/main/java/org/sonar/plugins/checkstyle/metadata/CheckUtil.java
+++ b/src/main/java/org/sonar/plugins/checkstyle/metadata/CheckUtil.java
@@ -25,10 +25,7 @@ import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
-import com.puppycrawl.tools.checkstyle.PackageNamesLoader;
-import com.puppycrawl.tools.checkstyle.PackageObjectFactory;
 import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
-import com.puppycrawl.tools.checkstyle.api.CheckstyleException;
 import com.puppycrawl.tools.checkstyle.checks.javadoc.AbstractJavadocCheck;
 import com.puppycrawl.tools.checkstyle.utils.JavadocUtil;
 import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
@@ -38,7 +35,7 @@ public final class CheckUtil {
     }
 
     public static String getModifiableTokens(String checkName) {
-        final AbstractCheck checkResult = getCheck(checkName);
+        final AbstractCheck checkResult = ModuleFactory.getCheck(checkName);
         final String result;
         if (AbstractJavadocCheck.class.isAssignableFrom(checkResult.getClass())) {
             final AbstractJavadocCheck javadocCheck = (AbstractJavadocCheck) checkResult;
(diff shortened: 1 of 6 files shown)


muhlba91 (approved):

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckUtil.java
```diff
@@ -60,11 +62,19 @@ else if (AbstractCheck.class.isAssignableFrom(checkResult.getClass())) {
...
     private static AbstractCheck getCheck(String checkName) {
-        final ClassLoader classLoader = CheckstyleMetadata.class.getClassLoader();
+        if (packageObjectFactory == null) {
+            try {
+                final ClassLoader classLoader = CheckstyleMetadata.class.getClassLoader();
+                final Set<String> packageNames = PackageNamesLoader.getPackageNames(classLoader);
+                packageObjectFactory = new PackageObjectFactory(packageNames, classLoader);
+            }
+            catch (CheckstyleException ex) {
+                throw new IllegalStateException(ex);
```
romani, line 72: During wrapping of exception we should provide context of execution in this case it is checkname

romani (changes requested):
Items

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckUtil.java
rnveach: There is no extra "context" to provide.

The only parameter of the method, `checkName`, is not used in this try/catch so it has no basis on the exception. The only thing to know is what method did the call and caused the exception, and that is already part of the stack trace.

Any exception that does occur, all information on the issue will be from the original exception.

rnveach:
@romani ping

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckUtil.java
romani: IllegalStateException("exception happened during loading of " + checkName, ex);

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckUtil.java
```diff
@@ -34,6 +34,8 @@
 import com.puppycrawl.tools.checkstyle.utils.TokenUtil;
 
 public final class CheckUtil {
+    private static PackageObjectFactory packageObjectFactory;
```
romani, line 37: Util class should stay stateless.
Let's move this factory to class that calls getCheck

romani (changes requested):
Items

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckUtil.java
rnveach: Done

src/main/java/org/sonar/plugins/checkstyle/metadata/CheckUtil.java
rnveach: Done.

romani (approved):
Ok to merge

PR #514, opened 2024-04-08
misc: Remove duplicate scopes

Bananeweizen:
The API expects a list of unique scopes there.


Bananeweizen:
d77b04fe37c50127ad13d664cd8595642df80390 replaced the deprecated module scope by project scope. However, project scope is listed twice now, and the duplicates should rather be removed.

PR #515, opened 2024-04-09
infra: avoid maven warnings

Bananeweizen:
Use the release property instead of source and target for compilation.


Bananeweizen:
That avoids these warnings: 
grafik

Also remove the property `java.version` as that is the name of a pre-defined property injected into the maven properties, containing the actual JDK runtime version (like `21.0.1`). It's really confusing to override that injected property, and could potentially lead to errors at other locations.

Issue #587, opened 2026-04-27
upgrade to checkstyle 13.4.1


romani:
https://checkstyle.org/releasenotes.html#Release_13.4.1

muhlba91:
sonarcloud PR: https://github.com/SonarSource/sonar-update-center-properties/pull/897

romani:
@muhlba91 , thanks a lot!!

PR #588, opened 2026-04-27
feat: dependency updates

muhlba91:
this updates all possible dependencies up to the version where they don't break the plugin.

we still need to resolve the sonarcloud analysis.

it also upgrades checkstyle w.r.t. #587.

it also resolves errors found in the project through recreating the sonarcloud project (see #589 for reference).

proof that the compiled plugin is working:
Screenshot 2026-04-27 at 6 10 39 PM


muhlba91:
note: merging since no errors are shown and to check #589.

PR #600, opened 2026-05-08
feat: add config option to disable the sensor (fixes #601)

### LINKED ISSUE #601: Disabling the plugin should be supported
We got reports of the **checkstyle** binary crashing on parsing java files during executing the plugin sensor, and that crashes the whole analysis as well. 

I have just done a quick test where I have created a new QP with only one active rule from the built-in (non-checkstyle) Java rules. Still, when running sonar-scanner on the client the checkstyle sensor was executed. So if I am correct, having no active checkstyle rules in the QP does not prevent the sensor execution on its own. Rule filtering seems to be performed post-check, on the server side.

I have created a PR that adds a config option `sonar.checkstyle.enabled`.
This option makes it possible to completely disable the checkstyle sensor for a certain project while the plugin is still available globally.

The config option defaults to true so there will be no need for any config modifications for existing users.

As other plugins also implement **_enabled_** or _**skip**_ config options, I thought that this config option would be a convenient way to work around these occasional checkstyle crashes.

Magyarimiki:
This PR adds a config option `sonar.checkstyle.enabled`.

While using the plugin there were some cases when **checkstyle** was unable to parse tricky Java code. This resulted in the whole analysis failing.

The above option makes it possible to completely disable the checkstyle sensor for a certain project while the plugin is still available globally.

The config option defaults to **true** so there will be no need for any config modifications for existing users.

This PR fixes #601.

### DIFF
--- a/src/main/java/org/sonar/plugins/checkstyle/CheckstyleSensor.java
+++ b/src/main/java/org/sonar/plugins/checkstyle/CheckstyleSensor.java
@@ -19,11 +19,15 @@
 
 package org.sonar.plugins.checkstyle;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.sonar.api.batch.sensor.SensorContext;
 import org.sonar.api.batch.sensor.SensorDescriptor;
 import org.sonar.api.scanner.sensor.ProjectSensor;
 
 public class CheckstyleSensor implements ProjectSensor {
+    private static final String CHECKSTYLE_ENABLED = "sonar.checkstyle.enabled";
+    private static final Logger LOG = LoggerFactory.getLogger(CheckstyleSensor.class);
 
     private final CheckstyleExecutor executor;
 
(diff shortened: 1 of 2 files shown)


muhlba91:
thank you for the PR.

however, in my opinion, this should be handled via different profiles on the sonar side.
if a project does require different checks and plugins (due to whatever reason), there should be a different profile created and associated with the corresponding project(s).
in such a new profile, checkstyle rules can be excluded.

Magyarimiki:
Thanks for your feedback.

I believe if we have the checkstyle plugin installed on the server and we run the analysis on a java project, the checkstyle 
sensor will still execute even if zero rules are activated in the quality profile. Of course, none of the issues will be shown on the UI with no active rules.

We got reports of  the checkstyle binary crashing on parsing java files, and that crashes the whole analysis as well. Even with no active checkstyle rules the sensor ran and crashed. 

I have just done a quick test where I have created a new QP with only one active rule from the built-in  (non-checkstyle) Java rules. Still, when running **sonar-scanner**  on the client the checkstyle sensor was executed. So if I am correct, having no checkstyle rules in the QP  does not prevent the sensor execution on its own.

As other plugins also implement **enabled** or **skip** config options, I thought that this config option would be a convenient way to fix this problem for our users.

muhlba91:
thanks for the testing and describing the problem in those details.
could you please create a small issue describing this (copy-paste your comment and PR description is sufficient?
then reference it in this PR and we are good to go. :)

muhlba91 (approved):

